PR #6, opened 2023-01-04
format security key validation message to match govuk design system

Adam-C-Mitchell:
Reformatting of the validation message that appears when an error occurs related to two-factor authentication via a security key (at /two-factor-webauthn) to match the error summary component from the GOVUK Design System, stipulated here: https://design-system.service.gov.uk/components/error-summary/.


quis:
See also https://github.com/alphagov/notifications-admin/pull/4506

jonathan-owens-gds (approved):

PR #13, opened 2023-03-02
Add area label to area display lists

Adam-C-Mitchell:
Add label to area display lists to clarify the purpose of the element to screenreader users.

### DIFF
--- a/app/templates/views/broadcast/partials/dashboard-table.html
+++ b/app/templates/views/broadcast/partials/dashboard-table.html
@@ -27,14 +27,16 @@
           {{ item.starts_at|format_date_human|title }} at {{ item.starts_at|format_time }}
         </p>
       {% endif %}
-      <ul class="area-list">
-        {% for area in item.areas %}
-          {% if loop.first %}
-            <p class="govuk-!-margin-right-2">Area: </p>
-          {% endif %}
-          <li class="area-list-item area-list-item--unremoveable area-list-item--smaller">{{ area.name }}</li>
-        {% endfor %}
-      </ul>
+      <div>
+        <p class="govuk-!-margin-right-2 area-list">Area: </p>
+        <ul class="area-list">
+          {% for area in item.areas %}
+            {% if loop.first %}
+            {% endif %}
+            <li class="area-list-item area-list-item--unremoveable area-list-item--smaller">{{ area.name }}</li>
+          {% endfor %}
+        </ul>
+      </div>
     </div>
   </div>
   {% if loop.last %}
--- a/app/templates/views/broadcast/partials/dashboard-table.html
+++ b/app/templates/views/broadcast/partials/dashboard-table.html
@@ -31,8 +31,6 @@
         <p class="govuk-!-margin-right-2 area-list">Area: </p>
         <ul class="area-list">
           {% for area in item.areas %}
-            {% if loop.first %}
-            {% endif %}
             <li class="area-list-item area-list-item--unremoveable area-list-item--smaller">{{ area.name }}</li>
           {% endfor %}
         </ul>
--- a/app/templates/views/broadcast/partials/dashboard-table.html
+++ b/app/templates/views/broadcast/partials/dashboard-table.html
@@ -29,6 +29,9 @@
       {% endif %}
       <ul class="area-list">
         {% for area in item.areas %}
+          {% if loop.first %}
+            <p class="govuk-!-margin-right-2">Area: </p>
+          {% endif %}
           <li class="area-list-item area-list-item--unremoveable area-list-item--smaller">{{ area.name }}</li>
         {% endfor %}
       </ul>
--- a/app/templates/views/broadcast/preview-areas.html
+++ b/app/templates/views/broadcast/preview-areas.html
@@ -28,6 +28,7 @@
   {% for area in broadcast_message.areas %}
     {% if loop.first %}
       <ul class="area-list">
+        <p class="govuk-!-margin-right-2">Area: </p>
     {% endif %}
         <li class="area-list-item">
           {{ area.name }}
--- a/app/templates/views/broadcast/preview-areas.html
+++ b/app/templates/views/broadcast/preview-areas.html
@@ -27,23 +27,25 @@
 
   {% for area in broadcast_message.areas %}
     {% if loop.first %}
-      <ul class="area-list">
-        <p class="govuk-!-margin-right-2">Area: </p>
+      <div>
+        <p class="govuk-!-margin-right-2 area-list">Area: </p>
+        <ul class="area-list">
     {% endif %}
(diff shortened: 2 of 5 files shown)


app/templates/views/broadcast/partials/dashboard-table.html
```diff
@@ -29,6 +29,9 @@ <h2>
       {% endif %}
       <ul class="area-list">
         {% for area in item.areas %}
+          {% if loop.first %}
+            <p class="govuk-!-margin-right-2">Area: </p>
+          {% endif %}
```
gds-gharper: I'm not sure `<p>` is an acceptable child element of a `<ul>`?

app/templates/views/broadcast/preview-areas.html
```diff
@@ -28,6 +28,7 @@
   {% for area in broadcast_message.areas %}
     {% if loop.first %}
       <ul class="area-list">
+        <p class="govuk-!-margin-right-2">Area: </p>
```
gds-gharper: Same here `<p>` isn't a valid child element of `<ul>`, perhaps make a new `<li class="area">` or move the `<p>` outside the `<ul>` ?

gds-gharper (changes requested):

app/templates/views/broadcast/preview-areas.html
Adam-C-Mitchell: Don't know why I missed that. Amended.

app/templates/views/broadcast/partials/dashboard-table.html
```diff
@@ -27,11 +27,16 @@ <h2>
...
-      <ul class="area-list">
-        {% for area in item.areas %}
-          <li class="area-list-item area-list-item--unremoveable area-list-item--smaller">{{ area.name }}</li>
-        {% endfor %}
-      </ul>
+      <div>
+        <p class="govuk-!-margin-right-2 area-list">Area: </p>
+        <ul class="area-list">
+          {% for area in item.areas %}
+            {% if loop.first %}
```
jonathan-owens-gds: Can this redundant if-block be removed?

gds-gharper (approved):

app/templates/views/broadcast/preview-areas.html
```diff
@@ -27,22 +27,25 @@
 
   {% for area in broadcast_message.areas %}
     {% if loop.first %}
-      <ul class="area-list">
+      <div>
+        <p class="govuk-!-margin-right-2 area-list">Area: </p>
+        <ul class="area-list">
     {% endif %}
```
gds-gharper: Why not just move this outside the for loop, then you can lose the loop.first

app/templates/views/broadcast/preview-areas.html
```diff
@@ -27,22 +27,25 @@
...
+                  <line stroke-width="1.75" id="svg_1" y2="0" x2="6" y1="12" x1="6" stroke="currentColor" fill="none"/>
+                  <line stroke-width="1.75" id="svg_3" y2="6" x2="0" y1="6" x1="12" stroke="currentColor" fill="none"/>
+                </g>
+              </svg>
+            </a>
+          </li>
     {% if loop.last %}
-      </ul>
+        </ul>
+      </div>
```
gds-gharper: The indendation here looks off, also I think this would be simpler just to move outside the loop.

gds-gharper (changes requested):

app/templates/views/broadcast/preview-areas.html
Adam-C-Mitchell: I know we never have any scenarios in which an area is not provided - but the reason behind this is to only create an ul if there are objects in the array. Otherwise, we'd have an empty list, and a p tag that says 'Area: ' without any areas. If there ever were, hypothetically, a scenario in which no areas were provided, it would make more sense to have the if statement, than to remove it, from an accessibility standpoint

app/templates/views/broadcast/preview-areas.html
Adam-C-Mitchell: The indentation has been bumped because there's an additional enclosing div to keep the elements inline.

app/templates/views/broadcast/preview-areas.html
Adam-C-Mitchell: To be fair, this was something in the code previously and I had just assumed that that was the rationale behind it - happy to remove if you still don't think it makes sense

gds-gharper: You could wrap the whole thing in an if if you want that:

```
{% if broadcast_message.areas %}
  <div>
    <p class="govuk-!-margin-right-2 area-list">Area: </p>
    <ul class="area-list">
      {% for area in broadcast_message.areas %}
      {% endfor %}
    </ul>
  </div>
{% endif %}
```

app/templates/views/broadcast/preview-areas.html
gds-gharper: No, I mean I think the endif for that loop.last is way out from the ending.  I wanted to highlight more lines but the code review wouldn't let me.  It'll probably look better if we sort out the loop.last loop.first things and move them out of the loop

Adam-C-Mitchell: Have tidied up now

app/templates/views/broadcast/preview-areas.html
```diff
@@ -25,31 +25,43 @@
...
+            </a>
+          </li>
+          {% if loop.last %}
+            {{ govukButton({
+              "element": "a",
+              "text": "Add another area",
+              "href": url_for('.choose_broadcast_library', service_id=current_service.id, broadcast_message_id=broadcast_message.id),
+              "classes": "govuk-button--secondary govuk-!-margin-bottom-5"
+            }) }}
+          {% endif %}
```
gds-gharper: You don't need to do this.

Just put this below the `{% endfor %}`

```
        <li>
          {{ govukButton({
            "element": "a",
            "text": "Add another area",
            "href": url_for('.choose_broadcast_library', service_id=current_service.id, broadcast_message_id=broadcast_message.id),
            "classes": "govuk-button--secondary govuk-!-margin-bottom-5"
          }) }}
      </li>
```

Then remove the whole if

gds-gharper (changes requested):

app/templates/views/broadcast/preview-areas.html
Adam-C-Mitchell: Agree, except for it being contained in a list item

gds-gharper (approved):

app/templates/views/broadcast/partials/dashboard-table.html
Adam-C-Mitchell: Have removed

PR #19, opened 2023-03-10
Broadcast Form Back Link State Retention

Adam-C-Mitchell:
As part of an audit into accessibility issues with the admin application, it was identified that clicking the back link on the form for creating a new broadcast does not retain state as one would expect it to. When returning to the form for adding "reference" and "content" fields to the database, their respective values are lost, effectively restarting the creation of an alert.
This PR fixes this issue in the interest of accessibility and user experience, by passing the broadcast message id into the screen as a query parameter if applicable, and allowing those fields to be updated following initial creation (see https://github.com/alphagov/emergency-alerts-api/pull/12).

### DIFF
--- a/app/main/views/broadcast.py
+++ b/app/main/views/broadcast.py
@@ -135,22 +135,46 @@ def new_broadcast(service_id):
 def write_new_broadcast(service_id):
     form = BroadcastTemplateForm()
 
+    broadcast_message_id = request.args.get("broadcast_message_id")
+    broadcast_message = None
+
     if form.validate_on_submit():
-        broadcast_message = BroadcastMessage.create_from_content(
-            service_id=current_service.id,
-            content=form.template_content.data,
-            reference=form.name.data,
-        )
+        if broadcast_message_id:
+            BroadcastMessage.update_from_content(
+                service_id=current_service.id,
+                broadcast_message_id=broadcast_message_id,
+                content=form.template_content.data,
+                reference=form.name.data,
+            )
+        else:
+            broadcast_message = BroadcastMessage.create_from_content(
+                service_id=current_service.id,
+                content=form.template_content.data,
+                reference=form.name.data,
+            )
+            broadcast_message_id = broadcast_message.id
         return redirect(
             url_for(
                 ".choose_broadcast_library",
                 service_id=current_service.id,
-                broadcast_message_id=broadcast_message.id,
+                broadcast_message_id=broadcast_message_id,
             )
         )
 
+    if broadcast_message_id:
+        broadcast_message = BroadcastMessage.from_id(
+            broadcast_message_id,
+            service_id=current_service.id,
+        )
+        if broadcast_message.status == "draft":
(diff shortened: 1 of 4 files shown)


gds-gharper (changes requested):
Can we get some tests for this new functionality please?

app/models/broadcast_message.py
```diff
@@ -85,6 +85,17 @@ def create_from_content(cls, *, service_id, content, reference):
...
+    @classmethod
+    def update_from_content(cls, *, service_id, broadcast_message_id, content, reference):
+        broadcast_message_api_client.update_broadcast_message(
+            service_id=service_id,
+            broadcast_message_id=broadcast_message_id,
+            data={
+                "reference": reference,
+                "content": content,
+            },
+        )
```
gds-gharper, line 97: We should have a test to ensure that this gets called when appropriate.

app/main/views/broadcast.py
```diff
@@ -135,22 +135,46 @@ def new_broadcast(service_id):
...
-            reference=form.name.data,
-        )
+        if broadcast_message_id:
+            BroadcastMessage.update_from_content(
+                service_id=current_service.id,
+                broadcast_message_id=broadcast_message_id,
+                content=form.template_content.data,
+                reference=form.name.data,
+            )
+        else:
```
gds-gharper, line 149: Can we test this path in the unit tests please?

app/main/views/broadcast.py
```diff
@@ -135,22 +135,46 @@ def new_broadcast(service_id):
 def write_new_broadcast(service_id):
```
gds-gharper, line 135: Does it make sense to change the name of this function now?

write_new_broadcast could be a confusing name since it doesn't always create a new broadcast anymore

app/main/views/broadcast.py
gds-gharper, line 149: Could this be used with parameter changing to update a broadcast message which has already been sent, should we only allow them to update draft messages?

app/main/views/broadcast.py
Adam-C-Mitchell: I'd argue it should still be write new broadcast because it's not necessarily intended as being able to have create/edit capacity - the idea is just that whatever is entered into the form isn't lost if coming back to it from later on in the flow.

app/main/views/broadcast.py
Adam-C-Mitchell: Have checked for this and written tests for it - see the new commits

app/models/broadcast_message.py
Adam-C-Mitchell: See new commits - added tests to make sure update/create is called appropriately

app/main/views/broadcast.py
Adam-C-Mitchell: See new commits - added tests to make sure update/create is called appropriately

gds-gharper (approved):

PR #20, opened 2023-03-13
EAS-447 Containerized build image

jonathan-owens-gds:
As another step toward self-contained Fargate deployed microservices, the following changes have been made to the Emergency Alerts Admin repo:

- Consolidate RUN commands in Dockerfile to reduce layer count
- Reduce the session timeout from 20 hours to 30 minutes.

### DIFF
--- a/Dockerfile.eas-admin
+++ b/Dockerfile.eas-admin
@@ -8,9 +8,8 @@ ENV ADMIN_DIR=/eas/emergency-alerts-admin
 COPY . $ADMIN_DIR
 
 # Run emergency-alerts-admin
-RUN $PYTHON_VERSION -m venv $VENV_ADMIN
-RUN . $VENV_ADMIN/bin/activate && pip3 install wheel
-RUN cd $ADMIN_DIR && . $VENV_ADMIN/bin/activate && make bootstrap
+RUN $PYTHON_VERSION -m venv $VENV_ADMIN && . $VENV_ADMIN/bin/activate && \
+    pip3 install wheel && cd $ADMIN_DIR && make bootstrap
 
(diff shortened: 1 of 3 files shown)


simon-sorrell-gds-alerts (approved):
Looks fine to me. The only comment I would make is that I suspect the original comment you updated was wrong in the first place, as I believe this is an absolute session timeout, so it is based on 30 minutes from, the time of authentication, not 30 minutes of inactivity. The config variable name would also suggest that is the case. Either way it is still ok to proceed.

jonathan-owens-gds:
@simon-sorrell-gds-alerts I've updated the comment to indicate that the timeout is absolute

Dockerfile.eas-admin
```diff
@@ -8,9 +8,8 @@ ENV ADMIN_DIR=/eas/emergency-alerts-admin
 COPY . $ADMIN_DIR
 
 # Run emergency-alerts-admin
-RUN $PYTHON_VERSION -m venv $VENV_ADMIN
-RUN . $VENV_ADMIN/bin/activate && pip3 install wheel
-RUN cd $ADMIN_DIR && . $VENV_ADMIN/bin/activate && make bootstrap
+RUN $PYTHON_VERSION -m venv $VENV_ADMIN && . $VENV_ADMIN/bin/activate && \
```
lstevens-cabinet-office, line 11: This is fine to do in one RUN command but it will mean that it won't be able to cache the other layers (virtual env creation on pip installing etc). I think it may make the image slightly smaller though (but it increases build time) so its personal choice.

lstevens-cabinet-office (approved):

PR #21, opened 2023-03-15
Mar '23 tour pages redesign

Adam-C-Mitchell:
As part of an accessibility audit conducted into the admin application, the tour pages were raised as an improvement area. The following PR implements the new styling/asset changes to the tour, to make it more visually and functionally accessible.


simon-sorrell-gds-alerts (approved):
These updated entries look fine, so ok to go ahead.

gds-gharper (commented):
Is there any particular reason we got rid of the `<main>` tag and replaced it with a `<div>`.?  I would have thought this would decrease accessibility.

Adam-C-Mitchell:
> Is there any particular reason we got rid of the `<main>` tag and replaced it with a `<div>`.? I would have thought this would decrease accessibility.

Consistency, and also styling. The main tag could be added across all of the tour pages and restyled accordingly, but the one that it was removed on was the exception (and it also broke the page layout). The others used divs.

gds-gharper:
> > Is there any particular reason we got rid of the `<main>` tag and replaced it with a `<div>`.? I would have thought this would decrease accessibility.
> 
> Consistency, and also styling. The main tag could be added across all of the tour pages and restyled accordingly, but the one that it was removed on was the exception (and it also broke the page layout). The others used divs.

Ah, I had assumed that we'd used main everywhere (as a couple of the files I checked also had it).  Github search appears to be fairly useless for searching for xml/html tags (as I've just found out), however a quick grep through the repo locally shows only the following pages have it which isn't many considering.

```
app/static/error_pages/5xx.html
app/templates/fullwidth_template.html
app/templates/views/platform-admin/_base_template.html
app/templates/views/organisations/add-organisation.html
app/templates/views/broadcast/tour/live/2.html
app/templates/views/broadcast/tour/live/1.html
app/templates/views/broadcast/tour/6.html
app/templates/withnav_template.html
app/templates/org_template.html
app/assets/error_pages/5xx.html
app/assets/stylesheets/views/product-page.scss```

PR #28, opened 2023-03-30
EAS-856 Use role-based credentials

jonathan-owens-gds:
In container startup script, take on role-based credentials instead of short-lived token.

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -2,9 +2,24 @@
 
 echo "Start script executing for admin.."
 
+# Query the fixed container agent IP address for credentials (search the AWS docs
+# for "Task IAM role" for more information about this endpoint)
+session_tokens=$(curl 169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI)
+
+if [[ -z $CONTAINER_ROLE ]] || [[ "$CONTAINER_ROLE" == "" ]]; then
+  export CONTAINER_ROLE=$(echo $session_tokens | jq -j .RoleArn)
+fi
+
+function configure_container_role(){
+  aws configure set role_arn $CONTAINER_ROLE
+  aws configure set credential_source EcsContainer
+  aws configure set default.region eu-west-2
+}
+
 function run_admin(){
   cd $ADMIN_DIR;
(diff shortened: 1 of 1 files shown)


lstevens-cabinet-office (approved):

scripts/start.sh
```diff
@@ -2,9 +2,24 @@
 
 echo "Start script executing for admin.."
 
+# Query the fixed container agent IP address for credentials (search the AWS docs
+# for "Task IAM role" for more information about this endpoint)
+session_tokens=$(curl 169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI)
```
gds-gharper, line 7: A few comments

What happens when these tokens expire?

According to https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html there's an expiration in the return value.

1. Once these tokens expire will the service need to be restarted?
2. Is that something we need to schedule?

We don't seem to use the `session_tokens` variable unless CONTAINER_ROLE isn't set, should we move that fetch inside the if statement below, that way we only fetch then if required.

lstevens-cabinet-office: **What happens when these tokens expire?**
As we have set the role and credential source in the default profile it will then resource new credentials from the credential source.

**Once these tokens expire will the service need to be restarted?**
Nope, its refetched like the way it does if you set your IAM credentials hardcoded into your aws profile.

**Is that something we need to schedule?**
Nope

**We don't seem to use the session_tokens variable unless CONTAINER_ROLE isn't set, should we move that fetch inside the if statement below, that way we only fetch then if required.**
We shouldn't add this into the If statement as other functions may require this later on, the session_token variable is not dependant on that role existing.

gds-gharper: That's a little disingenuous, we don't declare all variables as global variables just in case we might need them in another scope elsewhere in other code, so why would we do it here?

It's not exported, so nothing outside this script uses it and it slows down the script (for something which is literally unused in this case if CONTAINER_ROLE is set) and could in the future incur extra costs (as it's using extra cloud resources every time this is run).

PR #53, opened 2023-05-17
EAS 904 Decoupled UI MVP

Adam-C-Mitchell:
MVP for the decoupled UI. Note this doesn't actually strip out most of the original Notify pages, just adjusts the content for everything that there are internal links to such that it is cohesively Emergency Alerts-related. Privacy and Accessibility pages are not included in this PR as these are pending approval.


app/main/views/index.py
```diff
@@ -201,20 +197,20 @@ def get_started_old():
...
+@main.route("/using-emergency-alerts/who-its-for")
 def who_its_for():
     return redirect(url_for(".who_can_use_notify"), 301)
 
 
-@main.route("/using-notify/who-can-use-notify")
+@main.route("/using-emergency-alerts/who-can-use-notify")
 def who_can_use_notify():
     return render_template(
         "views/guidance/who-can-use-notify.html",
```
jonathan-owens-gds, line 216: This page is still called `/who-can-use-notify.html` ?  (I presume it's a page that will be culled later... just checking)

app/templates/views/performance.html
```diff
@@ -115,7 +115,7 @@ <h2 class="govuk-heading-m" id='messages-sent-within-10-seconds'>
     </p>
   </div>
 
-  <h2 class="govuk-heading-m" id='organisations-using-notify'>
+  <h2 class="govuk-heading-m" id='organisations-using-emergency-alerts'>
     Organisations using Notify
```
jonathan-owens-gds: Change to "Organisations using Emergency Alerts"

app/main/views/index.py
Adam-C-Mitchell: I can change it - that page isn't actually ever displayed in the application (unless going to the page directly via URL, which is probably fine for now)

app/templates/views/security.html
```diff
@@ -1,89 +1,93 @@
...
-    <li>terms of use</li>
-    <li>approach to data sharing</li>
-  </ul>
+    <h2 class="heading-medium" id="data">Data</h2>
+    <p class="govuk-body">On Emergency Alerts, data is encrypted when:</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>it passes through the service</li>
+      <li>it’s stored on the service</li>
+    </ul>
+    <p class="govuk-body">Any recipient data you upload is only held for 7 days.</p>
```
jonathan-owens-gds: I don't think we need to refer to uploading recipient data

app/templates/views/security.html
```diff
@@ -1,89 +1,93 @@
...
+    <p class="govuk-body">If you <a class="govuk-link govuk-link--no-visited-state" href="/features/email#send-files">send a file by email</a>, the file will be available for the recipient to download for up to 78 weeks (18 months).</p>
```
jonathan-owens-gds: We also don't need to refer to sending a file by email

app/main/views/index.py
jonathan-owens-gds: It's probably safe to leave it for now

app/templates/views/security.html
```diff
@@ -1,89 +1,93 @@
...
+    <ul class="govuk-list govuk-list--bullet">
+      <li>compliance with National Cyber Security Centre (NCSC) Cloud Security Principles</li>
+      <li>protective monitoring to record activity, and raise alerts about any suspicious activity</li>
+      <li>using JSON Web Tokens, to avoid sending API keys when your service talks to Emergency Alerts</li>
+    </ul>
 
-  <h3 class="heading-small">Protect sensitive information</h3>
-  <p class="govuk-body">Some messages include sensitive information like security codes or password reset links.</p>
-  <p class="govuk-body">If you’re sending a message with sensitive information, you can choose to hide those details on the Notify dashboard once the message has been sent. This means that only the message recipient will be able to see that information.</p>
+    <h3 class="heading-small">Protect sensitive information</h3>
```
jonathan-owens-gds: This section (the 'h3' and the two following 'p' elements) should probably be removed - users won't be uploading sensitive PII to the system.

app/templates/views/security.html
```diff
@@ -1,89 +1,93 @@
...
-  </ul>
-  <p class="govuk-body">This approach also applies to the service providers Notify uses to send messages.</p>
+    <h2 class="heading-medium" id="information-risk-management">Information risk management</h2>
+    <p class="govuk-body">Our approach to information risk management follows NCSC guidance. It assesses:</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>how Emergency Alerts is built</li>
+      <li>the infrastructure Emergency Alerts is built upon</li>
+      <li>support for the Emergency Alerts service</li>
+    </ul>
+    <p class="govuk-body">This approach also applies to the service providers Emergency Alerts uses to send messages.</p>
```
jonathan-owens-gds: We don't use service providers directly - we could change this to something like
"This approach also applies to the Notify service that the Emergency Alerts service uses to send emails and SMS messages for account set-up and two-factor authentication."

app/templates/views/security.html
```diff
@@ -1,89 +1,93 @@
...
-  <p class="govuk-body">You can use Notify to send messages classified as ‘OFFICIAL’ or ‘OFFICIAL-SENSITIVE’ under the <a class="govuk-link govuk-link--no-visited-state" href="https://www.gov.uk/government/publications/government-security-classifications">Government Security Classifications</a> policy.</p>
-  <p class="govuk-body">Notify does not process data classified as ‘SECRET’ or ‘TOP SECRET’.</p>
-  <p class="govuk-body">The Notify team has Security Check (SC) level clearance from <a class="govuk-link govuk-link--no-visited-state" href="https://www.gov.uk/government/organisations/united-kingdom-security-vetting">United Kingdom Security Vetting</a> (UKSV).</p>
+    <h2 class="heading-medium" id="classifications-and-security-vetting">Classifications and security vetting</h2>
+    <p class="govuk-body">You can use Emergency Alerts to send messages classified as ‘OFFICIAL’ or ‘OFFICIAL-SENSITIVE’ under the <a class="govuk-link govuk-link--no-visited-state" href="https://www.gov.uk/government/publications/government-security-classifications">Government Security Classifications</a> policy.</p>
```
jonathan-owens-gds: I don't think this 'p' applies to EA - potentially remove

app/templates/views/security.html
```diff
@@ -1,89 +1,93 @@
...
+    <p class="govuk-body">Emergency Alerts does not process data classified as ‘SECRET’ or ‘TOP SECRET’.</p>
```
jonathan-owens-gds: I don't think this 'p' applies to EA - potentially remove

app/templates/views/using-emergency-alerts.html
```diff
@@ -13,18 +13,25 @@
...
 
-    <p class="govuk-body">The information on this page has moved.</p>
+    <p class="govuk-body">This information is for teams using GOV.UK Emergency Alerts to send alerts.</p>
 
-    <p class="govuk-body">Find out more about:</p>
+    <h2 class="heading-medium">Help and guidance</h2>
     <ul class="govuk-list govuk-list--bullet">
       <li><a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.trial_mode_new') }}">trial mode</a></li>
       <li><a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.message_status') }}">message status types</a></li>
       <li><a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.features_email') }}">email replies</a></li>
```
jonathan-owens-gds, line 24: the links for email, sms and letters don't apply to EAS - probably remove

mike-mcmaster (approved):

PR #78, opened 2023-08-01
Eas 1342 security key registration content fix

Adam-C-Mitchell:
Replacing instances of "Notify" with "Emergency Alerts" for the security key flow, plus fix of the error message appearing on screen even when there was a success by splitting out errors into conditional error summary components.


Adam-C-Mitchell:
Re-ran the PR checks because they got a 500 from the Node software repo. That’s happened a few times recently, strange...

simon-sorrell-gds-alerts (approved):
These are fairly minor changes, fine to go ahead.

PR #85, opened 2023-08-15
EAS-1327 Configurable Alert Duration

Adam-C-Mitchell:
This PR addresses the need to capture the intended duration of an alert from the Admin UI. See the corresponding PR in the API repo for further information: https://github.com/alphagov/emergency-alerts-api/pull/77
This change adds a step to the sending journey, after the preview of areas, that allows for the selection of the intended duration of the alert, i.e.,:
- 30 mins
- 3 hours
- 6 hours
- 22 hours

These duration values are formatted and displayed alongside any preview of the alert (wherever you're able to see the alert areas) - only if a duration is present (i.e., old alerts won't have specified a duration).
Additionally, this duration value is used to calculate the finishes_at time of the broadcast_event, instead of basing this on the channel, as was the case previously.


simon-sorrell-gds-alerts (approved):
These changes look fine. I do have one question though, which is more around the logic than anything. If I requested an alert with a duration of the next 6 hours, but the pending alert wasn't approved for 2 hours, would the duration of the broadcast now be set to 4 hours? I suspect it would still be 6 hours, but would this now be what I intended when I asked for 6 hours, expecting it from when I asked for it (if that makes sense)

Adam-C-Mitchell:
> These changes look fine. I do have one question though, which is more around the logic than anything. If I requested an alert with a duration of the next 6 hours, but the pending alert wasn't approved for 2 hours, would the duration of the broadcast now be set to 4 hours? I suspect it would still be 6 hours, but would this now be what I intended when I asked for 6 hours, expecting it from when I asked for it (if that makes sense)

Yeah, I had a chat with Chris about this - for now, if you say 6 hours for example, it'll be 6 hours from the point at which the alert is approved. From the API side, we're storing an attribute in the broadcast_message table called "duration" - so we're not actually calculating the finishes_at time until it's approved, at which point it just adds the duration to the current time. We may change this in future, but that's what we went for as a first pass.

In future I think we'll probably either want to clarify this somewhere when selecting duration, or when previewing duration we specify something like "6 hours from time of approval" - something to that effect

PR #89, opened 2023-09-04
Allow no alert duration specified; use old defaults

Adam-C-Mitchell:
This PR makes it such that no alert duration can be specified; therefore allowing any cases where it hasn't been specified and is 0 (or None, for whatever reason) is handled appropriately by just using the old approach of checking which channel the alert will be sent over. One use case for this would be when alerts are created directly via the API.


simon-sorrell-gds-alerts (approved):
I think it is a good idea to have this ultimate failsafe to prevent issues if the values are not set. We may need to re-evaluate this further down the line in case we decide on an alternative approach (rejecting if it is not set maybe). For now though, this is perfect.

PR #94, opened 2023-09-22
Initial removal of Notify content

Adam-C-Mitchell:
This PR captures the changes to the code to remove instances of Notify and replace with Emergency Alerts. This doesn't (yet) capture the removal of redundant content.

Critical changes so far:
- “NotifyModules” has been renamed to “EmergencyAlertsModules” across the application (i.e., the class itself and any references to it). It seems to be working properly, with all unit tests passing.
- “notify_client” (As well as NotifyAdminAPIClient and other clients used by the admin application) have been renamed to use Emergency Alerts (following respective naming conventions)
- CSS class names
- “data-notify-module” attributes have been renamed to “data-emergency-alerts-module”.
- CSS classes containing “notify” have been renamed to emergency alerts
- All instances of “Notify” in HTML text has of course been replaced with Emergency Alerts (grammar considered - I haven’t noticed anything off)
- The package files (package.json and package-lock.json) have been updated such that the package reflects Emergency Alerts rather than Notify.

Considerations:
- `pytest.ini`: Instances of Notify could be replaced here. However, as this is a test config file, it needs some investigation as to the implications of just changing the word “notify” to “emergency-alerts”. It seems safe, but we need to be sure.
- `config.py`: The config.py file itself has been left for now. Most of the environment variables prefixed with “NOTIFY_” have been changed to “EMERGENCY_ALERTS_”, but some - in particular “NOTIFY_ENVIRONMENT”, have been left. This is because “NOTIFY_ENVIRONMENT” will also need to have its name changed in environment variables passed in.
- the /docs/static-assets.md file needs changing if we’re keeping it. It displays how static assets are served but the PNG still references Notify.
- Everything under /paas-failwhale hasn’t been changed because I imagine there is no reason for us to be removing this - and if we are keeping it for historical reasons, it seems more sensible to just leave everything as is from the Notify days.
- /scripts/run_app_paas.sh also hasn’t been changed, for the same reasons as above.
- /manifest.yml.j2 hasn’t been changed because it can probably be deleted later, like the other paas artefacts - it looks like this is an artefact from the paas days and there’s nothing we need to retain about this but worth double-checking on.
- The README of course needs to be updated once we’re happy with everything.

The Makefile hasn’t been changed. This is one of the things that will need a closer review as we obviously still use it.


Adam-C-Mitchell:
PR checks currently failing because this depends on changes being made to utils - a PR has been raised for this and should be merged to main shortly. That should allow the failing tests to pass once done.

PR #112, opened 2024-02-09
EAS 1800 Changed Crown symbol

leylayaltiligil:
Updated govuk-frontend to 4.8.0 & govuk-frontend-jinja to 2.8.0, added useTudorCrown parameter into govukHeader to display new image and updated svgs as directed by govuk-frontend versioning guidance.

### DIFF
--- a/app/assets/error_pages/5xx.html
+++ b/app/assets/error_pages/5xx.html
@@ -1,133 +1,119 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template ">
-  <head>
-    <meta charset="utf-8" />
-    <title>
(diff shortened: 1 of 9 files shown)


Adam-C-Mitchell (approved):
This looks good to me. The version bump was necessary to display the adjusted crown symbol via the design system, as per https://github.com/alphagov/govuk-frontend/releases/tag/v4.8.0, and any instances of the manually-provided SVG updated to reflect the same.

gds-gharper:
I would prefer if we commit this pull request and deploy it with useTudorCrown: false, and then have a PR which changes it to true ready to go on the day.  That makes it a much smaller change and gives us time to bed in

gds-gharper (approved):

PR #116, opened 2024-02-23
Eas 1799 Add Robots.txt

AlecAshmore:
Created robots.txt alongside a new 'metadata_blueprint'.


gds-gharper (commented):
Mostly this looks good however I'd suggest a slightly different approach.  I don't think we need a dynamic page for this so I would create a static blueprint rather than a metadata one that way we can use it for other similar things and it reduces the overhead of the system when generating the page.

https://flask.palletsprojects.com/en/3.0.x/blueprints/

```
static = Blueprint('admin', __name__, static_folder='static')

@static.route('/robots.txt')
```

then you can simply have a plain text robots.txt in the static folder which already exists.  The test changes should work unchanged I believe.

jonathan-owens-gds (approved):

PR #120, opened 2024-03-08
Eas 1799 Add Robots.txt

AlecAshmore:
Created robots.txt with a static route


app/main/views/index.py
```diff
@@ -35,6 +44,11 @@ def privacy():
     return render_template("views/privacy.html")
 
 
+@main.route("/robots.txt")
+def robots():
```
gds-gharper, line 53: I would just call this static() rather than robots().  That way when we get other files which may need to be sent this way you can just add the `@main.route("/other-file")` and the name is still accurate

gds-gharper (changes requested):
There appear to be a lot of changes here unrelated to the robots.txt change, are these from a different branch or a different ticket you're working on.

It's also possible you've pulled in someone elses changes from another branch, if that's the case I can work with you to rebase the branch and bring only your changes in.

gds-gharper (approved):

PR #123, opened 2024-03-18
Upgrade flake8 to version 7.0.0

jonathan-owens-gds:


gds-gharper (approved):
Looks like we don't use the incompatible change https://flake8.pycqa.org/en/latest/release-notes/7.0.0.html#backwards-incompatible-changes.

PR #127, opened 2024-03-26
EAS-1932 Implementing postcode search functionality and creation of custom areas

leylayaltiligil:
- Postcode library page added
- OS Code Point with Polygons data added to database
- Database now stored in S3 bucket
- Ability to search for postcode, and use centroid of postcode polygon and submitted radius to create new area, added
- Leveraging existing deletion functionality to clear areas from message


jonathan-owens-gds (approved):

app/models/broadcast_message.py
```diff
@@ -271,6 +272,26 @@ def add_areas(self, *new_area_ids):
...
+        if id not in self.area_ids:
+            simple_polygons = list(circle_polygon)
+            areas = {}
+            areas["ids"] = [id]
+            areas["names"] = [id]
+            areas["aggregate_names"] = []
+            if os.environ.get("IN_CICD"):
+                areas["simple_polygons"] = [
+                    [[round(coord, 10) for coord in polygon] for polygon in polygons] for polygons in simple_polygons
+                ]
```
jonathan-owens-gds, line 285: I think the triple-nested list comprehension here is at the limit of what's ok.  It's fine if the comprehension expression stays as simple as it is - any more complex than this and it'd probably be more readable if it was broken out into for-loops.

jonathan-owens-gds (approved):

.codepipeline/buildspec-admin-build.yml
```diff
@@ -13,6 +13,8 @@ phases:
       - COMMIT_ID=$(echo $COMMIT_ID | cut -c 1-7)
       - echo Build started on `date`
       - echo Building Docker image...
+      - chmod +x ./scripts/get-broadcast-areas-db.sh
+      - ./scripts/get-broadcast-areas-db.sh "./app/broadcast_areas" "${BROADCAST_AREAS_BUCKET_NAME}" "True"
```
gds-gharper, line 17: Minor but you could simply run `/bin/bash scripts/get-broadcast-areas-db.sh` and get rid of the chmod

app/broadcast_areas/create-postcode-areas-db.py
```diff
@@ -0,0 +1,212 @@
...
+                    continue
+            # We only care about the exterior of the polygon, not an
+            # holes in it that may have been created by fixing self
+            # intersection
+            fixed_polygon = Polygon(buffered.exterior)
+
+            # Make sure the polygon is now valid, and that we haven’t
+            # drastically transformed the polygon by ‘fixing’ it
+            assert fixed_polygon.is_valid
+            # assert isclose(fixed_polygon.area, shapely_polygon.area, rel_tol=0.001)
```
gds-gharper: Commented out code not needed anymore?

app/broadcast_areas/create-postcode-areas-db.py
```diff
@@ -0,0 +1,212 @@
...
+
+            print(f"{indent}Polygon {index + 1}/{len(polygons)} fixed!")  # noqa: T201
```
gds-gharper: Given the noqa happens a lot, it's probably worth turning this warning off for the entire file:

https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores

app/main/forms.py
```diff
@@ -2563,3 +2580,23 @@ class PlatformAdminSearch(StripWhitespaceForm):
...
+    postcode = PostcodeSearchField("Postcode")
+    radius = GovukDecimalField(
+        "Add radius",
+        param_extensions={
+            "classes": "govuk-input govuk-input--width-5",
+            "suffix": {"text": "km"},
+            "attributes": {"pattern": "[0-9]*"},
+        },
+        validators=[
+            NumberRange(min=0.099, max=38.001, message="Enter a radius between 0.1km and 38.0km."),
```
gds-gharper, line 2601: Is there a particular reason we're limiting to 38km?

app/main/forms.py
```diff
@@ -282,6 +285,20 @@ class GovukIntegerField(GovukTextInputFieldMixin, IntegerField):
...
+    pass
+
+
+class PostcodeSearchField(GovukTextInputFieldMixin, SearchField):
+    input_type = "Search"
+    param_extensions = {"classes": "govuk-input--width-10"}
+    validators = [
+        DataRequired(message="Enter a postcode."),
+        MustContainAlphanumericCharacters(),
+        Length(max=255, message="Service name must be 255 characters or fewer"),
```
gds-gharper: I think you left the message from somewhere else when it fails, also, can't we just check if it contains a valid postcode rather than just containing both letters and numbers since it's a very specific format.  Either with the giant proper postcode regex or a much simpler strip whitespace then check against something like`[a-z]{1,2}\d{1,2}[a-z]?\d[a-z]{2}`.  Having written that regex I'd probably just use a library for it like https://pypi.org/project/postcode-validator/.

app/main/forms.py
```diff
@@ -2563,3 +2580,23 @@ class PlatformAdminSearch(StripWhitespaceForm):
...
+        "Add radius",
+        param_extensions={
+            "classes": "govuk-input govuk-input--width-5",
+            "suffix": {"text": "km"},
+            "attributes": {"pattern": "[0-9]*"},
+        },
+        validators=[
+            NumberRange(min=0.099, max=38.001, message="Enter a radius between 0.1km and 38.0km."),
+            DataRequired(message="Enter a radius."),
+            Only2DecimalPlaces(),
```
gds-gharper, line 2603: I don't think you can enter with decimal places, since up above the attributes pattern above doesn't allow a decimal point?

app/main/views/broadcast.py
```diff
@@ -350,6 +441,7 @@ def choose_broadcast_sub_area(service_id, broadcast_message_id, library_slug, ar
     )
     if form.validate_on_submit():
         broadcast_message.add_areas(*form.selected_areas)
+        print(*form.selected_areas)  # noqa: T201
```
gds-gharper: Leftover debugging?

app/models/broadcast_message.py
```diff
@@ -271,6 +272,29 @@ def add_areas(self, *new_area_ids):
...
+            areas = {}
+            areas["ids"] = [id]
+            areas["names"] = [id]
+            areas["aggregate_names"] = [id]
+            if os.environ.get("IN_CICD"):
+                areas["simple_polygons"] = [
+                    [[round(coord, 10) for coord in polygon] for polygon in polygons] for polygons in simple_polygons
+                ]
+            else:
+                areas["simple_polygons"] = simple_polygons
```
gds-gharper: Is there a particular reason we're doing things differently for the tests in this file, it feels like we're masking a bug here.

app/models/broadcast_message.py
```diff
@@ -298,10 +322,26 @@ def _update_areas(self, force_override=False):
...
+            "names": [area.name for area in self.areas],
+            "aggregate_names": [area.name for area in aggregate_areas(self.areas)],
+            "simple_polygons": self.simple_polygons.as_coordinate_pairs_lat_long,
+        }
+
+        data = {"areas": areas}
+
+        # TEMPORARY: while we migrate to a new format for "areas"
+        if force_override:
+            data["force_override"] = True
```
gds-gharper: Possibly create a ticket for this and link to it in the comment, so that anyone who wants to remove it can see if it's been done or not.

app/templates/views/broadcast/partials/area-map-postcodes.html
```diff
@@ -0,0 +1,142 @@
...
+
+    {% endfor %}
+  {% endif %}
+
+
+  var mapElement = document.getElementById('area-list-map');
+  var grandparent = mapElement.parentNode.parentNode;
+  var isInDetails = grandparent.className.indexOf('govuk-details') !== -1;
+  var areas = document.querySelectorAll('.area-list > .area-list-item');
+  var trimRegExp = /^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g;
```
gds-gharper: What are we using this for, it feels unnecessary (since the server side should be able to handle it anyway), either way I think it's unusual enough to warrant a comment.

app/templates/views/broadcast/partials/area-map-postcodes.html
```diff
@@ -0,0 +1,142 @@
+<script src="{{ asset_url('javascripts/leaflet.js') }}"></script>
+<script>
+
+  var polygons = []
```
gds-gharper: I don't know about the rest of the code but modern javascript tends to use `let` and `const` rather than `var`.  Otherwise you end up creating global variables a lot.

gds-gharper (changes requested):

app/models/broadcast_message.py
leylayaltiligil: This is because the GitHub test runner is failing with a really minute difference in actual and expected coordinates. It seems like a floating point precision issue just within the GitHub runner instances and doesn't have any behavioural impact.

gds-gharper: This feels like something we should dig into and resolve, I'm not keen on having code in the libraries which bypasses code.  I think that you are right and it's a floating point issue and we can probably resolve it by either massaging the test data or by being more relaxed when we're inspecting the results inside the test test results.

app/main/forms.py
leylayaltiligil: The minimum value of 100m / 0.1km was driven by MNO requirements for the most effective transmission to get full penetration without risk of over subscription of cell towers. Too small an area might be difficult for their (MNO) systems to calculate which cell sites and towers to use, and 38km was driven by COBR as bigger than a city and the max side of alerting for non country size areas.

app/broadcast_areas/create-postcode-areas-db.py
leylayaltiligil: The area creation scripts have been referenced in flake8 file exclusions.

app/main/forms.py
leylayaltiligil: I've fixed this and created IsPostcode validator that validates using postcode regex.

.codepipeline/buildspec-admin-build.yml
leylayaltiligil: I tried with both and the pipeline in Hosted Development environment can only run the script with chmod +x command.

app/main/views/broadcast.py
leylayaltiligil: I've removed this line.

app/models/broadcast_message.py
leylayaltiligil: The test now loops through expected coordinates and actual and checks that they're within a tolerance of 1e-12 of each other. This is now passing within Github runner instances also, whereas it previously didn't without the rounding fix.

app/templates/views/broadcast/partials/area-map-postcodes.html
leylayaltiligil: Aria-label now created in jinja and both aria-label and static aria-description are set within area-map.html, so this has been removed from partials.

app/main/forms.py
gds-gharper: Can we add that as a comment?

app/broadcast_areas/create-postcode-areas-db.py
leylayaltiligil: Removed from current version.

app/models/broadcast_message.py
leylayaltiligil: Ticked created and linked in comment.

app/templates/views/broadcast/partials/area-map-postcodes.html
leylayaltiligil: I've removed all var references and replaced with appropriate const or let.

app/main/validators.py
```diff
@@ -226,3 +226,27 @@ def __call__(self, form, field):
...
+        self.message = message
+
+    def __call__(self, form, field):
+        if field.data and not re.match(self.regex, str(field.data)):
+            raise ValidationError(self.message)
+
+
+class IsPostcode:
+    regex = re.compile(
+        r"^([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9]?[A-Za-z])))) [0-9][A-Za-z]{2})$"  # noqa: E501
```
gds-gharper: This probably fails if they don't include the space between `SW1A 1AA`, you can also simplify it greatly by making it case insensitive (which turns `[A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2}` into `[a-z][a-hj-y][0-9]{1,2}]`

Personally though I would just use a library https://pypi.org/project/postcodes-uk/, tht way you can leave the unit testing to them and not worry about maintaining it

gds-gharper (changes requested):

app/main/validators.py
leylayaltiligil: Using postcode-validator now instead for regex for validation & for formatting postcodes.

app/main/validators.py
```diff
@@ -226,3 +227,23 @@ def __call__(self, form, field):
...
+
+    def __init__(self, message="Enter a value with 2 decimal places."):
+        self.message = message
+
+    def __call__(self, form, field):
+        if field.data and not re.match(self.regex, str(field.data)):
+            raise ValidationError(self.message)
+
+
+class IsPostcode:
```
gds-gharper, line 243: Why are you using the postcode regex instead of just calling `postcodes_uk.validate`?  

The below code seems a lot simpler
```
    def __call__(self, form, field):
        if field.data and not postcodes_uk.validate(str(field.data)):
            raise ValidationError(self.message)
```

Ah, I see, you used a different postcode validator to the one I suggested which doesn't have a validate function.  Comparing the two libraries only one of them still exists on github. https://github.com/GouthamSiddhaarth/postcode_validator vs https://github.com/marco-cardoso/postcodes_uk.  Ultimately I'll leave it up to you.

gds-gharper (approved):

app/main/validators.py
leylayaltiligil: I used the regex because with postcode-validator's validation it throws its own error message that I'd need to overwrite. Postcodes_uk had conflicting dependencies, it required an earlier version of pytest than we currently have.

PR #132, opened 2024-04-25
EAS-2030 Coordinate Mapping

leylayaltiligil:
- Coordinates added to library page
- Ability to search for area initially, and then add radius area
- Ability to search for coordinates in Latitude and Longitude format
- Ability to search for coordinates in Eastings and Northings format
- Areas now on map, for postcode and coordinates, as Leaflet Circles in preparation for functionality to be added


app/broadcast_areas/create-postcode-areas-db.py
```diff
@@ -92,7 +92,6 @@ def clean_up_invalid_polygons(postcode, polygons, indent="    "):
 
             # Make sure the polygon is now valid, and that we haven’t
```
jonathan-owens-gds: This comment looks like it no longer applies to any code - remove or move further down to where it may be relevant?

app/broadcast_areas/create-postcode-areas-db.py
```diff
@@ -136,7 +135,21 @@ def polygons_and_simplified_polygons(feature, name, postcode):
     # Check that the simplification process hasn’t introduced bad data
     for dataset in output:
         for polygon in dataset:
-            assert Polygon(polygon).is_valid
+            if not Polygon(polygon).is_valid:
+                buffered = Polygon(polygon).buffer(0)
+                if isinstance(buffered, MultiPolygon):
+                    for polygon in list(buffered.geoms):
+                        assert polygon.is_valid
```
jonathan-owens-gds: Here an invalid polygon will just generate an AssertionError - we should include some context to help debugging, such as `assert polygon.is_valid, "polygon failed validity test"

app/broadcast_areas/create-postcode-areas-db.py
```diff
@@ -136,7 +135,21 @@ def polygons_and_simplified_polygons(feature, name, postcode):
...
+                    multipolygons = [[[x, y] for x, y in polygon.exterior.coords] for polygon in list(buffered.geoms)]
+                    polygons = Polygons(multipolygons)
+                    full = polygons
+                    simplified = polygons.smooth.simplify
+                else:
+                    fixed_polygon = Polygon(buffered.exterior)
+                    assert fixed_polygon.is_valid
```
jonathan-owens-gds: Similar to above comment - add a little context in the error output

jonathan-owens-gds (approved):

app/utils/broadcast.py
```diff
@@ -0,0 +1,267 @@
...
+from app.broadcast_areas.models import CustomBroadcastArea
+from app.formatters import round_to_significant_figures
+from app.main.forms import (
+    CartesianCoordinatesForm,
+    DecimalCoordinatesForm,
+    PostcodeForm,
+)
+
+
+def create_coordinate_area_id(coordinate_type, first_coordinate, second_coordinate, radius):
```
gds-gharper: Is this really an ID, isn't it more of a name/slug?

If this is being used as an ID then most of the content is likely to be unchanging which is a pretty large waste of entropy space (And in fact wastage like this caused issues for an MNO when we submitted 10 Polygons as it made one of the strings the software they used as an ID overflow the string).

app/main/forms.py
```diff
@@ -2591,18 +2592,142 @@ class PostcodeForm(StripWhitespaceForm):
...
+        self.second_coordinate.validate(form)
+        return self.first_coordinate.validate(form) and self.second_coordinate.validate(form)
+
+
+class CartesianCoordinatesForm(StripWhitespaceForm):
+    """
+    This form contains the input fields for creation of areas using cartesian coordinates i.e. Eastings & northings.
+    There is additional validation post-submission of the form that checks whether the coordinate lies within
+    either the UK or the test polygons in Finland, and if both are false then an error is appended to the form.
+    """
```
gds-gharper: We're unlikely to use this for anything other than Eastings/Northings I'd be tempted just to call this EastingNorthingCoordinatesForm.  That way we also don't have to manually mentally translate between cartesian and eastings/northings in the rest of the codebase.

app/main/forms.py
```diff
@@ -2591,18 +2592,142 @@ class PostcodeForm(StripWhitespaceForm):
...
+        param_extensions={
+            "fieldset": {"legend": {"classes": "govuk-visually-hidden"}},
+            "hint": {"text": "Select one option"},
+            "items": [
+                {"hint": {"text": "For example, 51.503630, -0.126770"}},
+                {"hint": {"text": "For example, 530111, 179963"}},
+            ],
+        },
+        validators=[DataRequired(message="Select which type of coordinates you'd like to use")],
+    )
```
gds-gharper: I'd be tempted to use more literal values like latitude_longitude and easting_northing.  I'm not sure the text "Decimal Degrees) is helping anyone here, but perhaps our content designer has an opinion.

app/main/forms.py
```diff
@@ -2591,18 +2592,142 @@ class PostcodeForm(StripWhitespaceForm):
...
+class DecimalCoordinatesForm(StripWhitespaceForm):
+    """
+    This form contains the input fields for creation of areas using decimal coordinates i.e. Latitude & longitude.
+    For the first coordinate, latitude, to be valid the value entered must be numeric and between -90 and 90,
+    hence the NumberRange validator.
+    For the second coordinate, longitude, to be valid the value again must be number and between -180 and 180.
+    There is additional validation post-submission of the form that checks whether the coordinate lies within
+    either the UK or the test polygons in Finland, and if both are false then an error is appended to the form.
+    """
+
```
gds-gharper: In the same way as EastingsNorthings I think I'd just call this LatitudeLongitudeCoordinatesForm as even if we do allow people to use other decimal based coordinate systems we'd proabably want them to be explicitly stated.

app/templates/views/broadcast/partials/area-map-postcodes.html
```diff
@@ -1,66 +1,142 @@
...
+  function getCoordinateType() {
+    const pageUrl = window.location.href;
+    if (pageUrl.includes('cartesian')) {
+      return 'cartesian';
+    } else if (pageUrl.includes('decimal')) {
+      return 'decimal';
+    } else if (pageUrl.includes('postcodes')) {
+      return 'postcode';
+    }
+  }
```
gds-gharper: This feels like a fairly fragile way of determining what type of page we're on.  The partial already has dynamic content so why not let the server decide.  If you still needed it and didn't want to create dynamic javascript you could always create a small hidden field with an id which you could then fetch the value of (which you're going to need when you submit the form anyway).

app/templates/views/broadcast/partials/area-map-postcodes.html
```diff
@@ -1,66 +1,142 @@
 <script src="{{ asset_url('javascripts/leaflet.js') }}"></script>
```
gds-gharper: This partial feels misnamed now, as it no longer just provides postcodes.

app/templates/views/broadcast/search-coordinates.html
```diff
@@ -0,0 +1,112 @@
...
+  {{ page_title }}
+{% endblock %}
+
+{% block extra_stylesheets %}
+  {% include "views/broadcast/partials/area-map-stylesheets.html" %}
+{% endblock %}
+
+{% block extra_javascripts %}
+  {% include "views/broadcast/partials/area-map-postcodes.html" %}
+{% endblock %}
```
gds-gharper: In fact, if this is the only place where this partial is included then you should already have coordinate_type in a variable and can simply pass it in to the partial

gds-gharper (changes requested):

app/templates/views/broadcast/partials/area-map-postcodes.html
leylayaltiligil: I've removed this and instead use the coordinate_type variable passed in already for coordinates and for postcodes it checks for a variable only set when map is rendered on postcode page.

app/main/forms.py
leylayaltiligil: I've changed all references to cartesian and decimal coordinates to instead be 'easting_northing' and 'latitude_longitude', respectively, and have checked with the team regarding the need for "decimal degrees" to be there.

app/main/views/broadcast.py
gds-gharper, line 1: I think my main comment here isn't something we should fix right now, but we're really overloading this one view as can be evidenced by the fact that it's nearly 1000 lines of code.  Right now this one view performs 3 similar but quite different different functions (postcode selection, predefined area selection and coordinate selection).

I feel like this is overcomplicating the code with many if statements (repeatedly checking if the slug is postcode or coordinate etc) and it would be better to create completely different views for these user journeys.  We could still re-use code by putting it in libraries or components but it would make the actual code substantially simpler to read.

It would take quite a bit to re-work it now so it's probably outside the realm of this PR, but it should certainly be added as a technical debt ticket.

app/templates/views/broadcast/partials/custom-area-map-javascripts.html
gds-gharper, line 1: There's quite a bit of fairly difficult to unit test code in this file, would it make sense to move some of it into actual javascript modules and then call it with parameters?

app/utils/broadcast.py
gds-gharper, line 1: Shouldn't there be some unit tests code included in this PR for this library?

Otherwise that's quite a bit of new code which is only tested by incidental tests.

gds-gharper (changes requested):

app/main/views/broadcast.py
leylayaltiligil: I've created a ticket for this.

gds-gharper (approved):

PR #147, opened 2024-05-28
EAS-1577 Session Timeout

leylayaltiligil:
This PR consists of changes to the session, these being:

- Flask session length set to 60 mins
- Inactivity dialog appears after 28 mins of inactivity (last inactivity time is updated with every request and stored in local storage)
- Expiry dialog appears 2 mins before the flask session is due to expire to let users know
- Sets timings for dialogs to be displayed depending on whether or not the user is a specific functional test user


app/assets/javascripts/timeout.js
```diff
@@ -0,0 +1,149 @@
...
+const timeout_warning_mins = 58; // Minutes until popup displayed warning user of end of session
+
+(function (window) {
+  "use strict";
+
+  function isLoggedIn() {
+    // Checking whether user logged in or not based on whether logged_in_at exists
+    let timestamp;
+    if (typeof logged_in_at == 'undefined' || logged_in_at == 'None') {
+      timestamp =  'Test';
```
gds-gharper: I'm not sure having a magic string is a great way to determine that someone is logged in, couldn't you just return false here and remove the else clause?

That would also mean you wouldn't have to declare timestamp above and could declare it as const

app/assets/javascripts/timeout.js
```diff
@@ -0,0 +1,149 @@
...
+      timestamp = new Date(logged_in_at);
+    }
+    return (timestamp instanceof Date) & !isNaN(timestamp);
+  }
+  function loggedInAt() {
+    if (typeof logged_in_at == 'undefined') {
+      return new Date(); // This needs to be created for the tests
+    } else {
+      return new Date(logged_in_at);
+    }
```
gds-gharper: You don't need this else clause, you can simply `return new Date(logged_in_at)`

app/assets/javascripts/timeout.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  function getTimeLeftUntilExpiry() {
+    // calculates how long left in session, before final timeout, after each request
+    let expiry_time = new Date();
+    expiry_time.setMinutes(
+      loggedInAt().getMinutes() + timeout_warning_mins
+    );
+    let current_time = new Date();
```
gds-gharper: I'm not amazingly happy with doing date maths this way even though the documentation does say that values outside of 0-60 will update the date accordingly.  Perhaps consider using a proper date mathematic library like moment?

It would probably simplify quite a bit of the rest of the code too.

app/assets/javascripts/timeout.js
```diff
@@ -0,0 +1,149 @@
...
+    // Logs user out if no response when the inactivity popup shows
+    inactivity_logout = setTimeout(function () {
+      inactivity_popup.close();
+      signOutRedirect();
+    }, 1000 * 60 * initial_warning_mins);
+  }
+
+  function sessionExpiryPopup() {
+    // displays session expiry popup after timeout
+    const session_expiry_popup = document.querySelector("#expiry");
```
gds-gharper: Any particular reason you're using querySelector here rather than getElementById?  You're querying based on an id so we should try and be consistent

app/assets/javascripts/timeout.js
```diff
@@ -0,0 +1,149 @@
...
+    let current_page = window.location.pathname;
+    $.ajax("/sign-out", {
+      method: "GET",
+      success: function() {
+        window.location.href = "/sign-in?next="+current_page+"&status=inactive";
+      }
+    })
+      .fail((response) => {
+        console.log(response);
+      });
```
gds-gharper: Any particular reason you're using promises for fail and callbacks for success?

What happens if current_page contains characters which need URI escaping and/or it is a page which has been POSTed to or there are query string parameters?

package.json
```diff
@@ -49,7 +49,7 @@
     "gulp-sass": "5.0.0",
     "gulp-sass-lint": "1.4.0",
     "gulp-uglify": "3.0.2",
-    "jest": "29.0.1",
+    "jest": "^29.0.1",
```
gds-gharper: Is there a reason we've made this a little more strict about version updates?  If we're changing this would it be worth bringing us up to date with the current version 29.7?

tests/javascripts/timeout.test.js
```diff
@@ -0,0 +1,206 @@
...
+        "You can no longer extend your session"
+      );
+    } else {
+      expect(isDialogOpen(expiry_popup)).toEqual(false);
+      // not to be visible
+    }
+  });
+
+  test("Sign in again button should close dialog and sign out ", () => {
+    const expiry_button = document.getElementById(
```
gds-gharper: Again we seem to be using getElementById and getSelector (with an id passed) interchangeably, we should pick one and stick to it

gds-gharper (changes requested):
There are a lot of whitespace only changes here, for anyone else reason you can append `w=1` to the URL to hide the pure whitespace changes.

I think some of this code has been removed by refactors as well, so it would be worth rebasing against master in order to ensure you're up to date.

tests/javascripts/timeout.test.js
leylayaltiligil: I've removed all references to querySelector in favour of getElementbyId or getElementsByTagName for the selection of certain tags.

app/assets/javascripts/timeout.js
```diff
@@ -0,0 +1,186 @@
...
+    const inactivityTimeRemainingMsg = document.getElementById("time-remaining-message");
+    let timeLeft = Math.floor(getTimeLeftUntilSessionExpiry() / 60000);
+    let message;
+    if (timeLeft == 1) {
+      message = "You have 1 minute remaining in your session.";
+    } else if (timeLeft > 1) {
+      message = "You have "+timeLeft+" minutes remaining in your session.";
+    } else if (timeLeft < 1) {
+      message = "You have less than a minute remaining in your session.";
+    }
```
gds-gharper: I would argue one of these should be an else, rather than an else if

app/main/views/broadcast.py
```diff
@@ -124,7 +124,7 @@ def get_broadcast_dashboard_partials(service_id):
             "views/broadcast/partials/dashboard-table.html",
             broadcasts=broadcast_messages.with_status("pending-approval", "broadcasting"),
             empty_message="You do not have any current alerts",
-            view_broadcast_endpoint=".view_current_broadcast",
+            view_broadcast_endpoint="main.view_current_broadcast",
```
gds-gharper: Any particular reason this is being changed?

tests/app/main/test_permissions.py
```diff
@@ -38,7 +38,7 @@
         ("main.view_template", {"template_id": sample_uuid()}),
         ("main.view_template_versions", {"template_id": sample_uuid()}),
         ("main.view_template_version", {"template_id": sample_uuid(), "version": 1}),
-        ("no_cookie.view_letter_template_preview", {"template_id": sample_uuid(), "filetype": "pdf"}),
+        ("main.view_letter_template_preview", {"template_id": sample_uuid(), "filetype": "pdf"}),
```
gds-gharper: I think letter templates have been removed, you may wish to ensure that you're up to date with master/main

gds-gharper (changes requested):

jonathan-owens-gds (dismissed):

gds-gharper (dismissed):

jonathan-owens-gds (approved):

PR #159, opened 2024-06-28
EAS 2216 - Account Throttling

Adam-C-Mitchell:
This PR adds 'X-Forwarded-For' parameter to request as a header and adjusts existing tests accordingly.


app/notify_client/__init__.py
```diff
@@ -38,6 +38,7 @@ def _add_request_id_header(headers):
             return headers
         headers["X-B3-TraceId"] = request.request_id
         headers["X-B3-SpanId"] = request.span_id
+        headers["X-Forwarded-For"] = request.headers.get("X-Forwarded-For", "127.0.0.1")
```
gds-gharper: If there isn't an X-Forwarded-For then we shouldn't send in localhost, because it's not correct, simply don't include it if it's not forwarded for, it's certainly not correct to have it forwarded for localhost

gds-gharper (changes requested):

gds-gharper (approved):

PR #162, opened 2024-07-18
EAS-2209 Remove send.py module

jonathan-owens-gds:
Code removed:
- Module send.py & tests
- Module contact_list.py & tests
- UI elements related to sending docs to contacts


tests/app/main/views/test_templates.py
```diff
@@ -2152,25 +1989,25 @@ def test_should_show_hint_once_template_redacted(
...
-def test_should_not_show_redaction_stuff_for_letters(
-    client_request,
-    mocker,
-    fake_uuid,
-    mock_get_service_letter_template,
-    mock_get_template_folders,
-    single_letter_contact_block,
-):
-    mocker.patch("app.main.views.templates.get_page_count_for_letter", return_value=1)
+# def test_should_not_show_redaction_stuff_for_letters(
```
Adam-C-Mitchell: Is there a reason that this was commented out rather than outright removed?

app/main/forms.py
```diff
@@ -1248,36 +1245,36 @@ def validate_template_content(self, field):
         BroadcastLength()(None, field)
 
 
-class LetterAddressForm(StripWhitespaceForm):
-    def __init__(self, *args, allow_international_letters=False, **kwargs):
-        self.allow_international_letters = allow_international_letters
-        super().__init__(*args, **kwargs)
+# class LetterAddressForm(StripWhitespaceForm):
```
Adam-C-Mitchell: Same for some of the commented-out code in this file, shouldn't we just remove it? If we find we need it back we can pull it in from an older commit

Adam-C-Mitchell (approved):

PR #188, opened 2024-09-11
EAS-2278 Increasing password complexity

leylayaltiligil:
This PR includes:
- Updating both govuk-frontend and govuk-frontend-jinja to latest versions
- Resolved breaking changes & adjusted tests
- Adding pwdpy package for entropy calculations
- Adding LowEntropy validator for all password fields, except for where current password is required
- Adjusting unit tests by ensuring previously valid passwords are made more complex to retain validity with the forms


gds-gharper (changes requested):
just need to update the text for a low entropy password

PR #201, opened 2024-10-21
Updated html of the view-message for the alerts

AlecAshmore:


AlecAshmore:
Investigating why it's failed

gds-gharper (commented):
What is actually shown/given to a screen reader when we're using the id's instead of the names?

app/utils/broadcast.py
```diff
@@ -301,3 +301,20 @@ def get_centroid_if_postcode_in_db(postcode, form):
...
+    else:
+        return area_name
+
+
+def format_areas_list(areas_list):
+    return (
+        [format_area_name(area.name) for area in areas_list]
+        if (len(areas_list) == 1 or type(areas_list) is not CustomBroadcastAreas)
+        else [format_area_name(area) for area in areas_list.items]
+    )
```
gds-gharper: This is very awkward to read.  There's no real reason to try and cram everything onto one line and it just makes things harder to understand, I had to read this 3-4 times just to work out what was going on.

Also I suspect that if you have CustomBroadcastAreas with a length of one this will break because it'll revert to CustomBroadcastArea functionality but will still be trying to access an array.

I'd be tempted to
```
if type(areas_list) is CustomBroadcastAreas
    return [format_area_name(area.name) for area in areas_list]
return [format_area_name(area) for area in areas_list.items]
```
if you wanted to do belt and bracers you could

```
if type(areas_list) is CustomBroadcastAreas
    return [format_area_name(area.name) for area in areas_list]
else if type(areas_list) is CustomBroadcastAreas
    return [format_area_name(area) for area in areas_list.items]
# Either throw an error, an empty string or areas_list unchanged here
```

You also want unit tests for both of these new functions

gds-gharper (changes requested):

app/utils/broadcast.py
AlecAshmore: Adjusted the areas formatting and introduced unit tests on both the functions. 
I've kept the == 1 length as inside of the jinja it checks against this and only the first result is returned.

app/utils/broadcast.py
```diff
@@ -301,3 +301,22 @@ def get_centroid_if_postcode_in_db(postcode, form):
...
+    else:
+        return area_name
+
+
+def format_areas_list(areas_list):
+    if isinstance(areas_list, CustomBroadcastAreas):
+        if len(areas_list) == 1:
+            return [format_area_name(area.name) for area in areas_list]
+        else:
+            return [format_area_name(area.name) for area in areas_list.items]
```
gds-gharper: This still doesn't work right?  What happens if you pass a CustomBroadcastAreas with 1 instance, it will still have to use items?

The only way you don't want to loop through the items is if it's a CustomBroadcastArea (singular, without an array).

This is why in my suggestion I said  to check against instance of CustomBroadcastAreas vs CustomBroadcastArea

gds-gharper (changes requested):

app/utils/broadcast.py
AlecAshmore: Removed the comparison to length being 1 and implemented the 'areas' v 'area' approach

gds-gharper (dismissed):
Looks good now, good tests

Adam-C-Mitchell (approved):

PR #218, opened 2025-02-03
EAS-2564 Draft alert functionality

leylayaltiligil:
This PR adds the Admin functionality for the drafting and updating of broadcast messages.
The main changes introduced in the PR are as below:

- Adds edit_broadcast route that updates the broadcast message fields depending on user interactions with the page and a banner that displayed
- Adds hidden fields to BroadcastTemplate form to store data for edit_broadcast route
- Adds check_can_update_status method to BroadcastMessage class and calls it within various routes to check whether broadcast message can be updated to requested status. If this results in an exception then this is handled and displayed on the page.
- Adds submit_broadcast_message route to allow submission of broadcast messages from view-message page
- Adds view_broadcast_versions route and relevant templates to display the different versions of the broadcast message
-  Adds replace_areas method to BroadcastMessage class to ensure that if a library area is to be added to the broadcast message, it can be but the current custom area is cleared as the two cannot be combined currently
- Moves and added routes to be under "current-broadcasts" category to ensure the correct side navigation link is highlighted
- Adds methods to broadcast_message_api_client to allow retrieval of broadcast message versions from database
- Adds and adjusts existing templates
- Adds utils functions created from repeated code within new routes
- Adjusts existing tests & adds new tests and fixtures


app/templates/views/broadcast/_broadcast_message_history.html
```diff
@@ -0,0 +1,33 @@
+<div class="govuk-grid-column-full alert-list-item">
+  <div class="govuk-grid-column-full">
+    <h2 class="message-name">{{ broadcast_message.reference }}</h2>
+    <p class="hint">
+      {% if broadcast_message.version > 1 %}
```
gds-gharper: Perhaps instead of checking the magic version number here you shoud create a function if "is_edited" which returns a true/false.  I'm also not keen on these manually incremented version numbers.

app/templates/views/broadcast/macros/enlarged-map.html
```diff
@@ -0,0 +1,55 @@
+{% macro enlarged_map(estimated_area, estimated_area_with_bleed, count_of_phones, count_of_phones_likely, is_custom_broadcast, areas) %}
```
gds-gharper, line 1: I assume this is from another story where we wanted a larger popup of the map?  Is it worth making this PR larger/longer just in case there are issues with this functionality?

app/templates/views/broadcast/partials/dashboard-table.html
```diff
@@ -2,7 +2,7 @@
 {% from "govuk_frontend_jinja/components/tag/macro.html" import govukTag %}
 
 <div class="ajax-block-container js-mark-focus-on-parent" data-classes-to-persist="js-child-has-focus">
-{% for item in broadcasts|sort|reverse|list %}
+{% for item in broadcasts|list %}
```
gds-gharper, line 5: I like that we're letting the controller determine the order here rather than the view

app/templates/views/new-password.html
```diff
@@ -35,7 +35,7 @@ <h1 class="heading-large">
                 "text" : form.password.errors[0] | safe
           } if form.password.errors[0] else None,
         }) }}
-        {{ page_footer("Continue") }}
+        {{ page_footer("Continue", button_name="continue") }}
```
gds-gharper: I think these are showing up from Jonathans PR, it should make it smaller when he merges later on

gds-gharper (changes requested):

app/templates/views/broadcast/macros/enlarged-map.html
leylayaltiligil: As discussed, will consider this with future PRs but will keep in this instance.

app/models/broadcast_message.py
```diff
@@ -420,6 +441,19 @@ def clear_areas(self, force_override=False):
...
+    def get_versions(self):
+        return broadcast_message_api_client.get_broadcast_message_versions(self.service_id, self.id)
+
+    def get_count_of_versions(self):
+        return len(self.get_versions())
+
+    def get_latest_version(self):
+        return (
+            broadcast_message_api_client.get_broadcast_message_versions(self.service_id, self.id)[0]
+            if len(broadcast_message_api_client.get_broadcast_message_versions(self.service_id, self.id)) > 0
```
gds-gharper: Would this not make identical api calls back to back?  Would it not be better to fetch the result from the API and then make decisions on that result rather than calling get_broadcast_message_versions twice?

gds-gharper (changes requested):

gds-gharper (dismissed):

Adam-C-Mitchell (approved):

PR #220, opened 2025-02-21
EAS-2569: Admin: Introduce platform admin approval for inviting, adding permissions, or API key creation

jamesorlakin:
This PR overhauls the invitation, edit permission, and API key views to adopt the new AdminAction entity from https://github.com/alphagov/emergency-alerts-api/pull/190 and provide an admin approval process.

This is most certainly an MVP, so all nits are very much welcomed!

## Overview

There are now two permissions which are considered 'sensitive' - `create_broadcasts` and `approve_broadcasts`. Inviting a user or editing a user to add such a permission will instead create an AdminAction instead. Inviting a user without those permissions works as normal (which I think is the intention, but the acceptance criteria doesn't really read that way) and permissions can always be removed without approval.

image
image

Platform admins may reach the 'Admin approvals' screen to approve or reject such requests. This will update the status of the AdminAction, and then actuate whatever function the old screens invoked anyway (e.g. actually invite the user, etc). The API itself has little involvement here from a business logic perspective but that seems to be the pattern from what I can tell (such as the admin client creating events).

image

API keys are also in scope here but are a little different. _All_ attempts to create an API key will create an AdminAction. It used to be that only members of a service with the `manage_api_keys` permission could reach the create page (being a `platform_admin` did not net any benefits) but I've relaxed that here. Service members or another platform admin can both propose creating a key, and then another platform admin must approve it. The platform admin that approves the key is the one that will get to see its secret value as a one-off.

### Points of note

- UI design is not my strong point by far, and I confess I haven't spent much time focussing on that aspect. I'm hopeful we can evolve it fairly quickly.
- I've been mirroring patterns from elsewhere in the codebase but I'm not fully 'grokked' yet - it's likely I've gone against the grain in some aspects.
- There's no mechanism to view prior requests, only currently pending ones. The data is technically in the DB though.
- ~~We do not check for duplicate requests.~~ We invalidate prior requests that are similar.
- ~~It's hardcoded to require approval from a different admin, but I wonder if we could relax this in non-prod. Same for team/test API keys?~~ We have a config option for that.
- It's not obvious to a user that there may be pending admin approvals. E.g. from the manage users screen they will see no pending invite, the API key screen will be empty, etc. We could have a little 'nudge' box that mentions there are pending approvals that an admin must action first.


app/main/views/manage_users.py
```diff
@@ -61,6 +67,7 @@ def invite_user(service_id, user_id=None):
                 "views/user-already-team-member.html",
                 user_to_invite=user_to_invite,
             )
+        # TODO: Check for pending admin action
```
gds-gharper: Should we handle this TODO before merging?

app/main/views/platform_admin.py
```diff
@@ -105,6 +114,44 @@ def platform_admin_services():
...
+    methods=["POST"],
+)
+@user_is_platform_admin
+def platform_review_admin_action(action_id, new_status):
+    action = admin_actions_api_client.get_admin_action_by_id(action_id)
+
+    if action["status"] != ADMIN_STATUS_PENDING:
+        flash("That action is not pending and cannot be reviewed")
+    elif new_status == "approved" and action["created_by"] == current_user.id:
+        # TODO: Do we turn this off for non-production?
```
gds-gharper: We should also decide upon this, though my gut says perhaps we should have a setting in the config, and developer environments can have that turned off.

tests/app/main/views/test_api_integration.py
```diff
@@ -107,25 +107,33 @@ def test_should_create_api_key_with_type_normal(
...
-        _expected_status=200,
+        _data={"key_name": key_name, "key_type": "normal"},
+        _expected_redirect=url_for("main.api_keys", service_id=SERVICE_ONE_ID),
     )
 
-    assert page.select_one("span.copy-to-clipboard__value").text == (
-        # The text should be exactly this, with no leading or trailing whitespace
-        f"some_default_key_name_12-{SERVICE_ONE_ID}-{fake_uuid}"
-    )
+    # TODO: How do we assert the flash in a redirect?
```
gds-gharper: Not sure, but we should definitely find out.  You could mock the flash function perhaps, but I suspect there's a better way

gds-gharper (changes requested):
Nice refactoring in here as well.

app/main/views/manage_users.py
jamesorlakin: Good question.

If the incoming request is identical we could just silently succeed and not touch anything. But I guess we have two choices around how we should handle similar but not identical requests (e.g. inviting a user but with different permissions). We could:
- Automatically reject the existing pending request and create a new one
- Error out the creation of a new request and notify the user there's already something pending.

Personally I'm drawn to the former, as a non-admin (e.g. a service member with `manage_service` permission) still gets to action their intent.

app/main/views/platform_admin.py
jamesorlakin: I'm drawn to having a config too. For my local needs I ended up just commenting this statement out!

tests/app/main/views/test_api_integration.py
jamesorlakin: Turns out you can grab it from the session variable and assert that

app/main/views/manage_users.py
jamesorlakin: I've created a new status of 'invalidated' which all similar or identical pending requests will transition to before it creates a new one.

app/main/views/platform_admin.py
jamesorlakin: I've added a simple config in bd07a98.

app/config.py
```diff
@@ -74,6 +74,8 @@ class Config(object):
     EXPIRY_WARNING_MINS = 58
     INACTIVITY_WARNING_DURATION = 2
 
+    ADMIN_ACTION_ALLOW_SELF_APPROVAL = True
+
     FUNCTIONAL_TEST_PERMANENT_SESSION_LIFETIME = 0.667 * 60
     FUNCTIONAL_TEST_INACTIVITY_MINS = 0.167
```
gds-gharper: We've had some major issues in the past with default on behaviour.  I would be tempted to default this to False and then pull it from the environment if it's set, or specifically set it to False only in the dev environment.

gds-gharper (dismissed):

app/config.py
jamesorlakin: Good thinking, I would prefer that too.

gds-gharper (approved):

PR #223, opened 2025-03-04
EAS-2569: Admin: Fix approve button being disabled for all platform admins

jamesorlakin:
I had a test to not allow self-approval, but seemingly not the inverse - and it focused on the endpoint logic rather than the rendered template.

This has been rectified and most of the time was spent battling BeautifulSoup to get it to actually select the button. I've resorted to a data attribute, but my understanding from the docs was something like `page.find_all("button", string="Approve")` should've worked, but it just matched all buttons?


leylayaltiligil (dismissed):

jonathan-owens-gds (dismissed):
Tested and verified in hosted-dev environment

jonathan-owens-gds (approved):

PR #232, opened 2025-03-29
EAS-2605 Service MNO Choices

jonathan-owens-gds:
- Services can now choose to broadcast to any combination of available networks.
- Existing services may also be modified to change the networks to which they are affiliated.


app/main/forms.py
```diff
@@ -1575,47 +1568,56 @@ def __init__(self, broadcast_channel, *args, **kwargs):
...
+        None,
+        choices=[
+            ("all", "All mobile networks"),
             ("ee", "EE"),
             ("o2", "O2"),
-            ("vodafone", "Vodafone"),
             ("three", "Three"),
-        ),
+            ("vodafone", "Vodafone"),
+        ],
```
gds-gharper, line 1579: Whilst this works, it might be more acceptable from a WCAG perspective to have a radio group and an accordion.

So you have a radio group with options which "All Networks" and "Select Networks".  Then then you click "Select Networks" it opens an accordion which shows the checkboxes for each of the operators.  That would likely save quite a bit of javascript as well, as it'll just be using standard govuk components.

It can work in the same way the sort and filter works here: https://www.gov.uk/search/all

jonathan-owens-gds: I've followed the wireframe in the ticket blob:https://gds-ea.atlassian.net/33711de0-9a7f-473c-a4e5-652feef2d7f3#media-blob-url=true&id=38137586-15a4-4e30-8bed-db7c27bef7cc&collection=&contextId=13063&width=1920&height=1080&alt=Screenshot%202025-03-06%20at%2016.39.24.png - does a change to this need to be approved by any stakeholders?

leylayaltiligil (approved):

PR #236, opened 2025-04-03
EAS-2547 redirect after password reset

leylayaltiligil:
This PR includes:

- Addition of redirect url, if password has just been reset, to take user to sign in page
- When password has been reset, redirect passes in reset_password parameter that ensures sign in page displays relevant text to user, explaining why they need to sign in again
- Adds test to assert text displayed on page correctly & adjusts existing tests


Adam-C-Mitchell (approved):
Looks good, approving on the proviso that the FTs are passing against it (which they are)

jonathan-owens-gds (approved):
Looks good

PR #252, opened 2025-05-08
EAS-2602 + EAS-2603: Admin: Introduce new unsigned XML downloading capability

jamesorlakin:
> **Requires https://github.com/alphagov/emergency-alerts-utils/pull/102**

This adds new download links to the alert's page next to geoJSON.

image

This naturally produces the unsigned XML we want in the desired format. It also uses the alert reference as a file name hint to the browser:
image

There is a degree of overlap code-wide, as the logic which creates the 'lambda event' dict lives in the API, but I've tried to re-use as many constants as I could.

CAP:
image

IBAG:
image

(Polygons snipped for brevity. IBAG uses a hardcoded message number of `00000001`)

### DIFF
--- a/tests/app/main/views/test_broadcast.py
+++ b/tests/app/main/views/test_broadcast.py
@@ -5545,7 +5545,7 @@ def test_can_get_geojson_simple(
             areas={
                 "ids": ["Bristol ID"],
                 "simple_polygons": [BRISTOL],
-                # We want the name in the JSON so make them different to the IDs for asserting``
+                # We want the name in the JSON so make them different to the IDs for asserting
                 "names": ["Bristol Name"],
             },
         ),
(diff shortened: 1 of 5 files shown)


Adam-C-Mitchell (dismissed):
Nice work.

tests/app/main/views/test_broadcast.py
```diff
@@ -5583,3 +5584,259 @@ def test_can_get_geojson_simple(
...
+            reference="Test name",
+            content="Test content",
+            areas={
+                "ids": ["Bristol"],
+                "simple_polygons": [BRISTOL],
+                "names": ["Bristol"],
+            },
+        ),
+    )
+    # TODO:
```
Adam-C-Mitchell: Did you mean to remove this?

jamesorlakin: Aha, yep!

Adam-C-Mitchell (approved):

PR #256, opened 2025-05-21
EAS-2654: Don't send Slack notifications if a request is from a functional test IP

jamesorlakin:
https://github.com/alphagov/emergency-alerts-infra/pull/1313 has been merged so preview should now feed this value in.


app/utils/admin_action.py
```diff
@@ -219,3 +231,12 @@ def _get_action_description_markdown(action_obj, action_service: Service):
         markdown = "Elevate themselves to become a full platform admin\n_(This request will automatically expire)_"
 
     return markdown
+
+
+def _is_request_from_functional_tests():
+    functional_test_ips = os.environ.get("FUNCTIONAL_TEST_IPS", "").split(",")
```
gds-gharper: Would it be better to put this in config https://github.com/alphagov/emergency-alerts-admin/blob/main/app/config.py and then fetch the value from the config rather than burying an environment fetch in a random other module?

app/utils/admin_action.py
```diff
@@ -131,6 +132,10 @@ def _admin_action_is_similar(action_obj1, action_obj2):
 
 
 def send_slack_notification(new_status, action_obj, action_service: Service):
+    if _is_request_from_functional_tests():
+        current_app.logger.info("Skipping sending SlackMessage because this was from functional tests", action_obj)
+        return
```
gds-gharper: Maybe call this function (and the variables around it) should_prevent_slack_alerting or something, then in the config where the IP's are denoted mark them as from the functional tests?

gds-gharper (dismissed):

app/utils/admin_action.py
jamesorlakin: I agree - in retrospect that was quite a silly approach but I blindly mirrored how API was doing it!

app/utils/admin_action.py
jamesorlakin: Hmm, I'm not too sure. You'd presumably save one line of code at best as you'll need to return for control flow? (Or could invert it and only run if the `if` passes, but I prefer happy control path and early return IMO)

E.g.
```python
def send_slack_notification(new_status, action_obj, action_service: Service):
    if should_prevent_slack_alerting():
        return
```

gds-gharper: It's mostly the name rather than saving code, so it can be used for other purposes rather than functional tests if required

jonathan-owens-gds (approved):

PR #257, opened 2025-05-27
EAS-2653: Admin: Create ZenDesk tickets for admin elevations when out of hours

jamesorlakin:
> **Requires https://github.com/alphagov/emergency-alerts-utils/pull/103 and https://github.com/alphagov/emergency-alerts-infra/pull/1317**

This will create Zendesk tickets (with the P1 flag to try and get it to PagerDuty) for any admin elevation event. As per the ticket the priority field is defined based on the the event: requesting to be an admin, someone approving it, or actually becoming one post-approval. This will only occur out of hours and will automatically change the priority and comment on an existing ticket if it's open.

This will only occur if enabled in the environment. There's a PR open above to add that to production.

Tickets do not get closed - I'm assuming it's up to the support folks to do so. The ZenDesk (as-is the Slack logic) is in the request thread, so with Zendesk's search and manipulation APIs having a latency of upwards of 1000ms this can produce notable page load lag. I would be tempted to have it happen asynchronously but:
1. It's probably not going to occur often enough
2. We don't really have a pattern for that (well, there's Celery but that's on the API-side) and I'm not that brushed up with asyncio in modern Python!

---

Examples (it's JSON because I'm using the API and don't have Zendesk account for the UI myself):

Initial ticket (we don't update the description, just add comments for subsequent events):
image
image

Priorities get changed and comments added over time:
image


jonathan-owens-gds (approved):
All looks good - approved

PR #265, opened 2025-07-01
EAS-2687 Add extra content to alert

leylayaltiligil:
This PR consists of the following:

- Addition of `choose-extra-content` template to display radio form enabling users to choose whether or not they'd like to add additional information to the alert
- Addition of `add-extra-content` template where users can enter into a textarea the additional content that they'd like to add to the alert
- Changes to and additions of functions pertinent to the ability to add, edit or remove `extra_content` from an alert
- Display of additional information on various pages
- Adjustments to the 'Return for edit' banner - now display in an error summary with any errors for missing alert attributes that are required
- Addition of relevant unit tests and required adjustments to existing


app/templates/components/alert-summary-list.html
```diff
@@ -134,13 +142,40 @@ <h2 id="enlarged-map-dialog-message" class="govuk-visually-hidden screenreader-c
...
+                },
+                'value': {
+                    'text': broadcast_message.extra_content | text_area_formatting if broadcast_message.extra_content else ""
+                },
+                'actions': {
+                    'items': [
+                    {
+                        'href': url_for('.add_extra_content', service_id=current_service.id, broadcast_message_id=broadcast_message.id),
+                        'text': conditional_extra_content_action_text,
+                        'visuallyHiddenText': "message",
```
jamesorlakin:
```suggestion
                        'visuallyHiddenText': "extra content",
```

app/templates/components/alert-summary-list.html
```diff
@@ -56,6 +56,16 @@ <h2 id="enlarged-map-dialog-message" class="govuk-visually-hidden screenreader-c
...
                 {{ broadcast_message.duration|format_seconds_duration_as_time }}
             </p>
+        {% else %}
+            {% if  current_service.broadcast_channel in ["test", "operator"] %}
+                <p class="govuk-!-margin-right-2 duration-preview">
+                    {{ 14400|format_seconds_duration_as_time }}
+                </p>
+            {% else %}
+                <p class="govuk-!-margin-right-2 duration-preview">
+                    {{ 81000|format_seconds_duration_as_time }}
```
jamesorlakin: I'm always a tad hesistant about magic numbers in codebases. I know we have this default set in `forms.py:1089` but I wonder whether it might be worth refactoring to a constant/config if we repeat it.

By no means a strict requirement though.

app/utils/broadcast.py
```diff
@@ -448,3 +480,24 @@ def redirect_dependent_on_alert_area(broadcast_message):
...
+
+def check_for_missing_fields(broadcast_message):
+    errors = []
+    edit_url = url_for(
+        ".write_new_broadcast",
+        service_id=current_service.id,
+        broadcast_message_id=broadcast_message.id,
+    )
+    if not broadcast_message.reference:
+        errors.append({"html": f"""<a href={edit_url}>Add alert reference</a>"""})
```
jamesorlakin: Would these URLs need to be quoted?

jamesorlakin (dismissed):
Nice work! Minor nits but nothing important

app/utils/broadcast.py
leylayaltiligil: The links worked without the URLs being quoted but have added quotes for consistency, as it also works.

app/main/forms.py
```diff
@@ -1017,6 +1017,26 @@ def use_template(self):
...
+        choices=[
+            ("yes", "Yes"),
+            ("no", "No"),
+        ],
+        param_extensions={
+            "fieldset": {"legend": {"classes": "govuk-visually-hidden"}},
+            "hint": {
+                "html": """<p class="govuk-body">This won't be sent to those receiving the alert,
+                but will be displayed as part of the alert on <a class="govuk-link govuk-link--no-visited-state"
+                href='https://www.gov.uk/alerts'>gov.uk/alerts</a>. </p>
```
gds-gharper: Any particular reason for the single quotes here as opposed to staying consistent with the double quotes?

app/utils/broadcast.py
gds-gharper: A lot of things work because browsers are pretty forgiving, but that doesn't mean we should do it.  In this case the standard does not require it, but more places on the web would suggest that it's damgerous not to and that you should always quote attributes (it also means they won't break if something randomly gets a space in it.

https://www.w3schools.com/html/html_attributes.asp

On top of this, it probably doesn't matter here, but you should also ensure that any of the quote marks you've chosen to use (single, double) are explicitly escaped (in this case with URL encoding, but in other cases you may have to choose other methods for escaping).  In this case you can ensure that the `url_for` function does not create unexpected characters so it's probably fine.  However problems like this add to the reasons you should not create html content in code and that it should always be created by your chosen templating engine (which should ensure that this is all html safe by default).

app/templates/components/alert-summary-list.html
leylayaltiligil: I've added the constants to config and replaced the values in `forms.py` and `alert-summary-list.html`.

app/main/views/broadcast.py
```diff
@@ -196,6 +199,159 @@ def new_broadcast(service_id):
...
+    if current_service.broadcast_channel == "operator":
+        """If the current service is an operator service, user is redirected
+        as extra_content attribute shouldn't be set for alerts created in operator services"""
+        return redirect(
+            url_for(
+                ".view_current_broadcast",
+                service_id=current_service.id,
+                broadcast_message_id=broadcast_message_id,
+            )
+        )
```
gds-gharper: This is the same section of code above in `choose_extra_content` copied.  Is it worth refactoring it to reduce replication?

app/templates/components/alert-summary-list.html
```diff
@@ -134,13 +142,40 @@ <h2 id="enlarged-map-dialog-message" class="govuk-visually-hidden screenreader-c
...
                 'visuallyHiddenText': "message",
             }
             ] if display_action else []
         }
-        },
+        }
+    ]
+    %}
+    <!-- The additional information row in the table will only be displayed if service isn't
+    an operator service, as the field isn't relevant to operator service alerts. -->
```
gds-gharper: I'm not a big fan of having HTML comments (which will be in the actual outputted HTML.  You can instead use handlebars comments `{{! }}`

gds-gharper (dismissed):

jamesorlakin (dismissed):

gds-gharper (approved):

jamesorlakin (approved):

PR #268, opened 2025-08-14
EAS-2644 Adding areas to Templates

leylayaltiligil:
This PR consists of the following:
- Addition of a model, `BaseBroadcast`, a generalised abstract class, created to avoid duplication of methods and properties in child classes `BroadcastMessage` & `Template`
- Addition of `Template` model & majority of template-related class methods moved from `Service` model to `Template` model
- Addition of `TemplateAPIClient` class and methods, some moved from `ServiceAPIClient`
- Addition of` broadcast_common.py`, containing views used by both broadcasts and templates functionality, with behaviour dependent on `message_type` (either **templates** or **broadcast**)
- Addition of templates views, containing functionality pertinent to the creation of Templates in different ways
- Adjustments wherever Template 'name' is referenced - attribute has been renamed "reference" to be more consistent and to allow for reduced code duplication between broadcast and template functionality
- Addition of `templateSummaryList` component for display on Template page
- Adjustments to styling of checkboxes - removing extra whitespace
- Addition of `_choose_template_fields.html`, corresponding form, `ChooseTemplateFieldsForm`, and functionality, allowing user to choose how to populate Templates
- Addition of tests asserting new functionality & adjustments to existing tests


app/assets/stylesheets/components/message.scss
```diff
@@ -54,7 +54,19 @@ a {
 .template-list {
   &-item {
     &-with-checkbox {
-      padding-left: $govuk-checkboxes-size + $govuk-checkboxes-label-padding-left-right;
+      padding-left: 44px + $govuk-checkboxes-label-padding-left-right;
```
gds-gharper: Why not just set the variable instead of hard coding the size?

app/broadcast_areas/utils.py
```diff
@@ -64,7 +64,7 @@ def _cluster_lower_tier_authorities(areas):
 
 def _aggregate_REPPIR_sites_by_local_authority(area):
     # Returns the area name, concatenated with the local authority
-    return f"{area.name}, {area.parent.name}"
+    return f"{area.name}, {area.parent.name}" if area.parent else f"{area.name}"
```
gds-gharper: Readability wise I'm not a huge fan of trailing if statements, especially with an else.  Mentally it means you have to unwind what you just read in order to understand it.

app/main/views/broadcast.py
```diff
@@ -39,6 +39,9 @@
 from app.models.broadcast_message import BroadcastMessage, BroadcastMessages
```
gds-gharper, line 34: It's almost impossible to tell what the actual changes are here.  It feels like the entire file has been shuffled around and refactored.  I think would have preferred if the reorganisation was accomplished in a separate PR to the actual changes.  I can't keep enough context in my head to understand this diff.

app/main/views/templates.py
```diff
@@ -94,10 +137,12 @@ def choose_template(service_id, template_type="all", template_folder_id=None):
     if request.method == "GET" and initial_state:
         templates_and_folders_form.op = initial_state
 
+    service = Service(service_api_client.get_service(service_id)["data"])
```
gds-gharper: Why is this not just current_service?  Other parts of this function are still referencing current_service

app/main/views/templates.py
```diff
@@ -329,10 +378,11 @@ def manage_template_folder(service_id, template_folder_id):
         )
         return redirect(url_for(".choose_template", service_id=service_id, template_folder_id=template_folder_id))
 
+    service = Service(service_api_client.get_service(service_id)["data"])
```
gds-gharper: Again, other parts of this function still reference current_service, why use a mixture of current_service and service?

app/main/views/templates.py
```diff
@@ -462,41 +534,41 @@ def edit_service_template(service_id, template_id):
...
                 ".action_blocked",
                 service_id=service_id,
-                notification_type=template["template_type"],
+                notification_type=template.template_type,
                 return_to="view_template",
                 template_id=template_id,
             )
         )
     else:
+        service = Service(service_api_client.get_service(service_id)["data"])
```
gds-gharper: Same question about service here

app/main/views/templates.py
```diff
@@ -607,12 +683,538 @@ def redact_template(service_id, template_id):
...
+                    service_id=service_id,
+                    template_type="broadcast",
+                    template_folder_id=template_folder_id,
+                )
+            )
+        elif template_fields == "area_only":
+            return redirect(
+                url_for(
+                    ".choose_library",
+                    service_id=current_service.id,
```
gds-gharper: Why does this use current_service.id when everything else in this function simply uses service_id?

app/main/views/templates.py
```diff
@@ -4,18 +4,58 @@
 from flask_login import current_user
 from notifications_python_client.errors import HTTPError
 
-from app import current_service, service_api_client, template_folder_api_client
+from app import (
```
gds-gharper, line 7: Should all of the routes in this be checking the user has permission to access the template/service?  I think they're all checking that the user has manage_templates functionality but I'm not sure they're all checking with `get_template_with_user_permission_or_403`

app/main/views/templates.py
```diff
@@ -607,12 +683,538 @@ def redact_template(service_id, template_id):
...
+            form=form,
+            back_link=url_for(
+                ".choose_library",
+                service_id=service_id,
+                message_id=template_id,
+                message_type="templates",
+                template_folder_id=template_folder_id,
+            ),
+        )
+
```
gds-gharper: We seem to be missing `service_id` and `current_service.id` we should try to be consistent.

app/main/views/templates.py
gds-gharper, line 1: Try to ensure consistent use of `service_id` and `current_service.id` through this file.

gds-gharper, line 1: There appears to be a lot of copied and pasted code from app/utils/broadcast.py in this file.  Just checking for the string `If they are within test or UK` shows 4 matches.

Could we consolidate some of it?

app/models/template.py
```diff
@@ -0,0 +1,129 @@
+from app.broadcast_areas.models import (
+    BroadcastAreaLibraries,
+    CustomBroadcastArea,
+    broadcast_area_libraries,
+)
+from app.broadcast_areas.utils import aggregate_areas, generate_aggregate_names
+from app.models.base_broadcast import BaseBroadcast
+from app.notify_client.template_api_client import template_api_client
+
+ESTIMATED_AREA_OF_LARGEST_UK_COUNTY = broadcast_area_libraries.get_areas(["lad23-E06000065"])[  # North Yorkshire
```
gds-gharper: This line seems to be copied from app/models/broadcast_message.py and app/models/base_broadcast.py I haven't checked but I suspect they may share more code, could the common functionality be consolidated?

app/utils/broadcast.py
```diff
@@ -511,9 +521,65 @@ def check_for_missing_fields(broadcast_message):
...
+
+def _get_choose_library_back_link(
+    service_id,
+    message_type,
+    message_id=None,
+    template_folder_id=None,
+):
+    if message_type == "broadcast":
+        return url_for(
+            ".choose_extra_content" if current_service.broadcast_channel != "operator" else ".write_new_broadcast",
```
gds-gharper: I'm not a huge fan of these magic string checks which are around.  Would it be better to have a method in the service model called "can_have_extra_content" or similar?

Also, still not a fan of trailing if/else as it means you end up having to read the code backwards.

gds-gharper (changes requested):

app/main/views/broadcast.py
leylayaltiligil: I've moved some functions back into their original places so the changes within the functions should now be easier to read.
The main changes to the file now are: 
- The removal of `write_new_broadcast` view function (moved to [broadcast_common.py](https://github.com/alphagov/emergency-alerts-admin/pull/268/files#diff-9fa30366fe4f5ec3455c1e0e0cb8364ec5e1fe9405a52cd6376c3a557a539598R34)) because of the adjustments necessary for Template creation
- Any instances of `Template` 'name' or 'template_content' replaced with 'reference' and 'content' respectively as part of the work to make the `Template` and `BroadcastMessage` flows more similar and consistent
- Did refactor `broadcast` function (now uses [update_broadcast](https://github.com/alphagov/emergency-alerts-admin/pull/268/files#diff-09dc9a05865349d00b85e381975b49e622d72d82a8fb17cbaeaba4594de0582aR1391) and [create_new_broadcast](https://github.com/alphagov/emergency-alerts-admin/pull/268/files#diff-09dc9a05865349d00b85e381975b49e622d72d82a8fb17cbaeaba4594de0582aR1430) views), just as adding Template changes did add to the code complexity here
- In many functions, `message` is passed into jinja templates as this allows for them to be used by both template and broadcast_message functionality and views
- `remove_postcode_area` & `remove_coordinate_area` consolidated into one `remove_custom_area_from_broadcast` view
- View function URLs updated to reflect that they now use broadcast view URLs and the "old" functions are used for both broadcast_message and templates, and then functionality depends on which type of message they are

app/main/views/templates.py
leylayaltiligil: I've adjusted so that service_id used whenever its been passed in to the function, otherwise the function uses current_service.

app/utils/broadcast.py
leylayaltiligil: I've added this property to the Service model and now reference this throughout.

app/main/views/broadcast.py
leylayaltiligil: Following discussion about this change, a lot of the views in this file have since been moved to `broadcast_common.py` and adjusted so they can be used for both BroadcastMessage and Template models. This means that a lot of the repeated functionality in their views has now been consolidated.

app/main/views/templates.py
leylayaltiligil: I've added a `from_id_or_403` method to both Template and BroadcastMessage models, that ensures the access and manage (or create for Broadcasts) permissions are checked against. Where these are not used, `current_user.has_template_folder_permission` checks users permissions instead, where necessary.

gds-gharper (approved):

PR #282, opened 2025-09-26
EAS-2834 Handle invalid area

leylayaltiligil:
This PR ensures that if an area is invalid, an error persists on the view/preview alert pages and will prevent submission of the alert.
This PR adds the `is_valid_area` method to `CustomBroadcastAreas` model and uses shapely to validate that the area is valid, this is then checked against in various routes.
The original issue - error page displayed when trying to access an alert with an invalid area, has been resolved by handling the exception (returning 0 for the model attribute if there's an exception) but to display the error the area is validated.

### DIFF
--- a/app/main/views/broadcast.py
+++ b/app/main/views/broadcast.py
@@ -40,6 +40,7 @@ from app.utils.broadcast import (
     format_areas_list,
     get_changed_alert_form_data,
     get_changed_extra_content_form_data,
+    get_error_message_from_topology_error,
     keep_alert_content_button_clicked,
     keep_alert_reference_button_clicked,
     overwrite_content_button_clicked,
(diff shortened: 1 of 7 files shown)


app/main/views/broadcast.py
```diff
@@ -537,6 +548,16 @@ def submit_broadcast_message(service_id, broadcast_message_id):
...
 
+    if type(broadcast_message.areas) is CustomBroadcastAreas and not broadcast_message.areas.is_valid_area():
+        errors = [
+            {
+                "text": "The area used is invalid and the alert cannot be sent. If the alert was created"
+                " through the API, report it to the alert creator. Otherwise report it to the "
+                "Emergency Alerts team."
+            }
+        ]
+        return render_current_alert_page(broadcast_message, hide_stop_link=True, errors=errors)
```
gds-gharper: Rather than repeating these lines is it worth sticking the text into a constant somewhere to be re-used?

Tht way if we have to update the message it's done in one place.

gds-gharper (changes requested):

gds-gharper (approved):

PR #289, opened 2025-10-09
EAS-2416 Add Flood Warning areas

leylayaltiligil:
This PR consists of the changes required to be able to select flood warning areas to add to the alert, using TA codes.
The changes in this PR are as follows:
- Addition of a script used to add flood warning target areas to SQLite db
- Addition of `is_flood_warning_target_area` property to `BroadcastArea` model
- Adjustments to `redirect_dependent_on_alert_area` function, to allow for greater reusability, now returns the url instead or the redirect of the url, so can be used more widely
- Adjusts `choose_library` view to account for added `Flood_Warning_Target_Areas` library
- Addition of `search_flood_warning_areas` view to render Flood Warning area selection page & handle addition of areas to alert
- Addition of `FloodWarningForm`
- Adjustment to `preview-areas` template to change link URL dependent on whether or not the alert has flood warning areas
- Addition of `search-flood-warning-areas` template
- Addition of relevant tests


app/broadcast_areas/models.py
```diff
@@ -62,6 +62,18 @@ def estimated_bleed_in_m(self):
...
 
+    @cached_property
+    def nearby_electoral_wards(self):
+        if not self.polygons:
+            return []
+        return broadcast_area_libraries.get_areas_with_simple_polygons(
+            [
+                # We only index electoral wards in the RTree
+                overlap.data
+                for overlap in rtree_index.query(Rect(*self.simple_polygons.bounds))
```
gds-gharper: I'm really not a big fan of these trailing statements (if/for etc).  It makes reading the code awkward.  In this case a variable randomly pops into scope and you don't find out what it is until you finish reading the line.

In general, I'm not a huge fan of code golf, it sacrifices readabilility for reduced size.

app/broadcast_areas/models.py
```diff
@@ -116,6 +128,15 @@ def count_of_phones(self):
...
         return self._count_of_phones or 0
 
+    @property
+    def estimated_count_of_phones(self):
+        return (
+            sum(
+                area.simple_polygons.ratio_of_intersection_with(self.polygons) * area.count_of_phones
+                for area in self.nearby_electoral_wards
+            )
+        ) or 0
```
gds-gharper: Same problem. here, we have 4 nested statements including a trailing loop.

app/broadcast_areas/models.py
```diff
@@ -228,6 +237,7 @@ def __init__(self, row):
         self.name_singular = name_singular
         self.is_group = bool(is_group)
         self.items = BroadcastAreasRepository().get_all_areas_for_library(self.id) if self.id != "postcodes" else []
+        self.item_ids = [item[0] for item in self.items]
```
gds-gharper: At least wish a zip the code reads left to right, even though it's not amazingly readable
```
self.item_ids = list(zip(*self.items))[0]
```

app/utils/broadcast.py
```diff
@@ -501,7 +501,7 @@ def redirect_dependent_on_alert_area(broadcast_message):
             message_type="broadcast",
         )
 
-    return redirect(redirect_url)
```
gds-gharper, line 504: This function no longer does what it's called.  I would change it to something like `get_alert_redirect_url` or `calculate_alert_redirect_url`

gds-gharper (changes requested):

app/broadcast_areas/models.py
leylayaltiligil: I've rewritten this as:
```
bounds = self.simple_polygons.bounds
search_rect = Rect(*bounds)  # Creates rectangle of the polygon bounds
overlaps = rtree_index.query(search_rect)  # We only index electoral wards in the RTree
electoral_ward_ids = [overlap.data for overlap in overlaps]  # extracts overlapping electoral ward IDs
return broadcast_area_libraries.get_areas_with_simple_polygons(electoral_ward_ids)
```

app/broadcast_areas/models.py
leylayaltiligil: I've rewritten this as:
```
total_estimated_phones = 0
for area in self.nearby_electoral_wards:
      # Iterate through electoral wards that overlap with polygon bounds and calculate phone count
      # using ratio of intersection and electoral ward's count of phones
      intersection_ratio = area.simple_polygons.ratio_of_intersection_with(self.polygons)
      estimated_phones = intersection_ratio * area.count_of_phones
      total_estimated_phones += estimated_phones
return total_estimated_phones or 0
```
(Indentation is correct in file, just difficult to do here when its not pasting nicely)

app/broadcast_areas/models.py
leylayaltiligil: I've rewritten this as:
```
self.item_ids = []
for item in self.items:
     # iterate through areas and append their IDs to list
      self.item_ids.append(item[0])
```

app/broadcast_areas/models.py
```diff
@@ -62,6 +62,18 @@ def estimated_bleed_in_m(self):
...
 
+    @cached_property
+    def nearby_electoral_wards(self):
+        # If area has polygons, we identify electoral wards that overlap the polygon bounds
+        if not self.polygons:
+            return []
+        bounds = self.simple_polygons.bounds
+        search_rect = Rect(*bounds)  # Creates rectangle of the polygon bounds
+        overlaps = rtree_index.query(search_rect)  # We only index electoral wards in the RTree
+        electoral_ward_ids = [overlap.data for overlap in overlaps]  # extracts overlapping electoral ward IDs
```
gds-gharper: There's still a trailing "for" in here.  In general if you have to work backwards through the line to understand what 's going on it's difficult to read.

app/broadcast_areas/models.py
```diff
@@ -228,6 +239,10 @@ def __init__(self, row):
         self.name_singular = name_singular
         self.is_group = bool(is_group)
         self.items = BroadcastAreasRepository().get_all_areas_for_library(self.id) if self.id != "postcodes" else []
+        self.item_ids = []
+        for item in self.items:
+            # iterate through areas and append their IDs to list
```
gds-gharper: Is this comment particularly helpful, I think the code reads exactly what the comment says

gds-gharper (changes requested):

app/broadcast_areas/models.py
Adam-C-Mitchell: I kind of disagree. It's Pythonic to have list comprehensions that way (https://peps.python.org/pep-0202/). Tbh, if you asked for my personal opinion, I prefer more imperative `for x in y` syntax as well (I'm a Java chap, old habits die hard) - but I don't think we should be discouraging what the Python style guide itself promotes as best practice.

gds-gharper: There's a nice discussion about it here https://www.reddit.com/r/Python/comments/26etjj/comment/chqlf49 and whilst I still don't like making people read right to left if offers some reasonable compromises.  There's some more discussion here which seems to agreee https://www.reddit.com/r/learnpython/comments/u0zq1w/is_list_comprehension_bad_code/

It feels like most people seem to agree that doing more than one thing (which some of the others were doing) makes it hard to read, but a simple assignment is probably ok,

Adam-C-Mitchell: I agree with that. A big part of Python's rather aggressive recommended limit on line length (79 as standard) is in an attempt to discourage people from nesting things inside of nests inside of nests (and cut cyclomatic complexity down). I guess my rule of thumb would be if it feels like you need brackets inside your list comprehension, don't use one.

gds-gharper (approved):

PR #297, opened 2025-11-17
EAS-2892: Resolving the incorrect area creation issue with Template `create_from_area` method

leylayaltiligil:
This PR resolves the issue in which the areas added to `Template` objects are incorrectly created. To resolve the issue, I've moved the area creation method from `BaseBroadcast` model into a file of utils functions, chosen to avoid circular dependency issues. The code for the relevant functions has just been transplanted across into another file to be referenced elsewhere, so the original comments have been retained too.
I've also adjusted existing unit tests and added some more for the utils functions.


app/models/broadcast_message.py
```diff
@@ -85,7 +85,7 @@ def create_from_area(cls, service_id, area_ids, template_id=None, content="", re
             "ids": area_ids,
             "names": [area.name for area in areas],
             "aggregate_names": aggregate_names,
-            "simple_polygons": [area.simple_polygons.as_coordinate_pairs_lat_long for area in areas],
+            "simple_polygons": [area.simple_polygons.as_coordinate_pairs_lat_long for area in areas][0],
```
jamesorlakin: Would this also need reflecting in `create_from_area` for templates?

gds-gharper (approved):

jamesorlakin (approved):

PR #318, opened 2026-01-15
EAS-2760 Delete draft alerts

jonathan-owens-gds:
Changes in this PR:

- Allow draft alerts to be discarded at the 'preview' stage
- Allow returned alerts to be discarded
- Let the user discard multiple or all draft alerts using checkbox selection


app/assets/javascripts/recalculateStickinessOnMutation.js
```diff
@@ -0,0 +1,28 @@
...
+
+    if (!targetElement)
+      return;
+
+    const observer = new MutationObserver((mutations) => {
+      mutations.forEach((mutation) => {
+          mutation.removedNodes.forEach(() => {
+              GOVUK.stickAtBottomWhenScrolling.recalculate();
+          });
+      });
```
gds-gharper, line 13: I tend to prefer to use `for (const mutation of mutations)` rather than jQuery's forEach these days.  It's built into the language and doesn't require lambdas.  There's a couple of other instances of this as well below.

gds-gharper (changes requested):

app/assets/javascripts/recalculateStickinessOnMutation.js
jonathan-owens-gds: Usage of 'forEach' replaced with 'for-of'

leylayaltiligil (dismissed):

app/assets/javascripts/discardDraftAlerts.js
```diff
@@ -0,0 +1,33 @@
...
+  "use strict";
+
+  const addDiscardDraftsClickHandler = function () {
+    const btn = document.getElementById("discard-alerts-button");
+    if (btn) {
+      btn.addEventListener("click", function () {
+        const checkboxes = document.querySelectorAll(
+          'input[type="checkbox"]:checked'
+        );
+        const checkedIds = Array.from(checkboxes).map((cb) => cb.id);
```
gds-gharper: You seem to be mixing `function() {}` and `() => {}` lambda syntax here.  it would be good to stick to one format.

app/assets/javascripts/discardDraftAlerts.js
```diff
@@ -0,0 +1,33 @@
...
+        if (checkedIds.length) {
+            const service_id = window.location.pathname
+                .split("/")
+                .filter(Boolean)[1];
+            fetch(`/services/${service_id}/discard-drafts`, {
```
gds-gharper: What happens if this call fails for any reason?  Is the user notified in any way or are they just going to keep waiting?

app/assets/javascripts/discardDraftAlerts.js
```diff
@@ -0,0 +1,33 @@
...
+                method: "POST",
+                headers: {
+                    "Content-Type": "application/json",
+                    "X-CSRFToken": btn.getAttribute("data-csrf-token"),
+                },
+                body: JSON.stringify({ draft_alerts: checkedIds }),
+            });
+        }
+        const newPath = window.location.pathname.replace('/select-drafts', '/current-alerts');
```
gds-gharper: Using `replace()` here means we're changing the current URL and explicitly avoiding putting it in the history.  This means back doesn't work as the user would expect it to.  Would it be better to use `assign()` here?

app/assets/javascripts/recalculateStickinessOnMutation.js
```diff
@@ -0,0 +1,28 @@
...
+  "use strict";
+
+  const addRecalculateStickinessOnMutation = function () {
+    const GOVUK = global.GOVUK;
+    const targetElement = document.querySelector('.ajax-block-container');
+
+    if (!targetElement)
+      return;
+
+    const observer = new MutationObserver((mutations) => {
```
gds-gharper: Again, we're mixing lambda styles here.  Would be good to stick to one.

app/assets/javascripts/recalculateStickinessOnMutation.js
```diff
@@ -0,0 +1,28 @@
+(function (global) {
+  "use strict";
+
+  const addRecalculateStickinessOnMutation = function () {
+    const GOVUK = global.GOVUK;
+    const targetElement = document.querySelector('.ajax-block-container');
+
+    if (!targetElement)
+      return;
```
gds-gharper: It's minor, but I tend to always put in braces {} on if statements, even for single instructions like this.  It just means if anyone ever decides to add something else before the return here they aren't suddenly surprised by the implicit scope change vs being explicit.

gds-gharper (changes requested):

app/assets/javascripts/discardDraftAlerts.js
jonathan-owens-gds: function format changed

app/assets/javascripts/discardDraftAlerts.js
jonathan-owens-gds: image
image
image
[CLICK "BACK"]
image

jonathan-owens-gds: The behaviour of back works ok here. Obviously it won't un-discard the messages, but it does return to the discard-alerts page

gds-gharper: I can't see the URL or the current history on those screenshots, but the documentation at https://developer.mozilla.org/en-US/docs/Web/API/Location/replace says it replaces your current entry in the history.  Perhaps current alerts was in there multiple times?
Happy to have a quick screen share to confirm

jonathan-owens-gds: Just to confirm: Working from a new tab (with no history)
image
image
image
[CLICK "BACK"]
image

jonathan-owens-gds: From the current-alerts page, the select-alerts page does show up in the history
image

app/assets/javascripts/discardDraftAlerts.js
jonathan-owens-gds: If any issue arises from calling 'fetch' on the given URL, I can throw the error so that it shows up in the console, but as with all calls to the `broadcast_message_api_client` in /app/models/broadcast_message.py, the calls are bare, i.e. not wrapped in any try-except handling or response checking. Any issues with the broadcast-message rejection step will be surfaced in the logging on the API instance, and the user will be aware of the failure because the alert will fail to be rejected.

app/assets/javascripts/recalculateStickinessOnMutation.js
```diff
@@ -0,0 +1,21 @@
...
+    const targetElement = document.querySelector('.ajax-block-container');
+    if (!targetElement) {
+      return;
+    }
+    const observer = new MutationObserver((mutations) => {
+      mutations.forEach((mutation) => {
+          mutation.removedNodes.forEach(() => {
+              window.GOVUK.stickAtBottomWhenScrolling.recalculate();
+          });
+      });
```
gds-gharper, line 13: There's still a couple of forEach here

tests/javascripts/recalculateStickinessOnMutation.test.js
```diff
@@ -0,0 +1,59 @@
...
+  test("called when '.keyline-block' removed from DOM", (done) => {
+    window.GOVUK.stickAtBottomWhenScrolling.init();
+
+    const recalculateSpy = jest.spyOn(window.GOVUK.stickAtBottomWhenScrolling, 'recalculate');
+
+    const keylineBlocks = document.querySelectorAll('.keyline-block');
+    const count = keylineBlocks.length;
+    keylineBlocks.forEach((block) => {
+        block.remove();
+    })
```
gds-gharper, line 52: And a forEach here

gds-gharper (changes requested):
Last 2 changes I think then it looks good

PR #333, opened 2026-02-12
EAS-2910 Filter & sort alerts

jonathan-owens-gds:
This PR adds filtering by alert status and sorting by date and alert reference to the following pages:

- Current alerts
- Past alerts
- Rejected alerts


app/main/views/broadcast.py
```diff
@@ -115,36 +138,49 @@ def delete_draft_alerts(service_id):
     )
 
 
-@main.route("/services/<uuid:service_id>/past-alerts")
+@main.route("/services/<uuid:service_id>/past-alerts", methods=["GET", "POST"])
 @user_has_permissions()
 @service_has_permission("broadcast")
 def broadcast_dashboard_previous(service_id):
+    page = "past-alerts"
+    filter, sort = _persist_and_return_filter_and_sort_values(page)
```
jamesorlakin, line 146: Small nit - filter is a Python built-in function which is being overridden here. It doesn't really matter because it's localised to these small endpoint handlers though.

app/main/views/broadcast.py
```diff
@@ -156,18 +192,71 @@ def broadcast_dashboard_updates(service_id):
...
+def _get_filter_options(broadcast_messages, filter=None, page="current-alerts"):
+    statuses = list(set(msg.status for msg in broadcast_messages))
+    filter_options = [{"value": status, "text": FILTER_TEXT[status]} for status in statuses]
+    if len(filter_options) > 1:
+        filter_options.insert(0, {"value": "none", "text": "All Alerts"})
+    return {"id": f"{page}-filter", "text": "Show", "options": filter_options, "selected": filter}
+
+
+def _get_sort_options(sort=None, page="current-alerts"):
+    sort_values = ["date-asc", "date-desc", "title-asc", "title-desc"]
```
jamesorlakin: Would this be more DRY, or probably not worth the hassle/overhead?

```suggestion
    sort_values = sorted(SORT_TEXT.keys())
```

jamesorlakin (dismissed):
Nice one

app/main/views/broadcast.py
jonathan-owens-gds: Probably a good idea - adding/removing sort keys only needs to happen at the top of the file

jamesorlakin (approved):

PR #338, opened 2026-02-18
EAS-2917: Include all polygons for an area when generating preview XML (CAP/IBAG)

jamesorlakin:
This fixes a misunderstanding in the data model used in admin for areas. The term area is a bit overloaded: in the CBC/XML logic, `areas` is a list of lists representing polygon coordinates, i.e. a flat list of shapes. In admin, `area` represents BroadcastArea objects, which could be single user-facing items (e.g. 'England') that actually represent more than one polygon (e.g. islands) when it comes to processing them.

The logic previously only used the first polygon returned by the Polygons class within the BroadcastArea object as it seemed like the surrounding list was superfluous, but now I correctly loop through every returned list of coordinate pairs and add them to the overall `areas` as CBC/XML would expect.

(It's in these scenarios I miss static typing!)

Since this is a bit confusing I've taken the liberty of making the test data consistent. Unless otherwise stated in the constant name it can be assumed to refer to only one polygon.

Anyhow, for the whole of the UK this means we end up with many areas rather than one per country on the left:
image


Adam-C-Mitchell (approved):
> It's in these scenarios I miss static typing!

Amen to that brother

PR #341, opened 2026-02-24
EAS-2943: manage templates improvements

Dave-Johnson-GDS:


jonathan-owens-gds (dismissed):

app/main/views/templates.py
```diff
@@ -70,6 +72,72 @@ def edit_template(service_id, template_id):
...
+@main.route("/services/<uuid:service_id>/templates/move-to", methods=["POST"])
+@main.route("/services/<uuid:service_id>/templates/folders/<uuid:template_folder_id>/move-to", methods=["POST"])
+@user_has_permissions(allow_org_user=True)
+def move_template(service_id, template_folder_id=None):
+
+    if not current_user.has_permissions("manage_templates"):
+        abort(403)
+
+    folders_to_move = json.dumps(request.form.getlist("templates_and_folders"))
+    if folders_to_move == "[]":
```
gds-gharper: Why are we serialising this to json to compare as a string, rather than just checking the length of the array returned by request.form.getlist("templates_and_folders")?

I feel like there's a better way to do this than throwing blobs of json around to templates.

gds-gharper (dismissed):

app/main/views/templates.py
Dave-Johnson-GDS: Thanks @gds-gharper - much more straight forward now without passing json around.

gds-gharper (approved):

PR #372, opened 2026-03-26
EAS-2878: Bulk add local authorities

leylayaltiligil:
This PR consists of the addition of the functionality to allow users to enter multiple local authorities (separated by newline) into text area, to then be added to an alert. This has already been introduced for Flood Warning Areas in this PR #326.
The PR consists of:
- Adjustments to `BroadcastAreaLibrary` model attributes, to allow for retrieving Local Authority ID using name and more efficient retrieval of all area names for library
- Addition of `LocalAuthorityBulkAreasForm` and validation of new textarea input field
- Addition of `search_local_authority_areas_as_a_list` view function, handling the addition of local authorities bulk input in text area
- Addition of [search-local-authority-list.html](https://github.com/alphagov/emergency-alerts-admin/pull/372/changes#diff-af466a78981183c66ebeb0eb9823d4d44eeb06c4dda93dc59fbeed9b74ccc113) page with textarea
- Adjustments to existing and addition of relevant tests


app/broadcast_areas/models.py
```diff
@@ -241,9 +241,14 @@ def __init__(self, row):
         self.name_singular = name_singular
         self.is_group = bool(is_group)
         self.items = BroadcastAreasRepository().get_all_areas_for_library(self.id) if self.id != "postcodes" else []
-        self.item_ids = []
-        for item in self.items:
-            self.item_ids.append(item[0])
+        self.item_ids = (
+            BroadcastAreasRepository().get_all_area_ids_for_library(self.id) if self.id != "postcodes" else []
+        )
```
gds-gharper: Is this not now calling `BroadcastAreasRepository().get_all_areas_for_library(self.id) if self.id != "postcodes" else []` twice?

app/broadcast_areas/repo.py
```diff
@@ -186,6 +186,31 @@ def get_all_areas_for_library(self, library_id):
...
+
+        return [row[0] for row in results]
+
+    def get_all_area_names_for_library(self, library_id):
+        # Countries don't have any children, so the above query wouldn't return anything.
+        q = """
+        SELECT id, LOWER(name)
+        FROM broadcast_areas
+        WHERE (broadcast_area_library_id = ?
+        AND id LIKE 'lad23-%')
```
gds-gharper, line 196: Is it worthwhile commenting this magic string?  I'm not sure why it's here in a very generically named function `get_all_area_names_for_library`

app/broadcast_areas/models.py
```diff
@@ -241,9 +241,14 @@ def __init__(self, row):
...
+        self.area_names_ids_lookup = (
+            BroadcastAreasRepository().get_all_area_names_for_library(self.id)
+            if self.id == "wd23-lad23-ctyua23"
```
gds-gharper: Worthwhile sticking a comment to tell poeple that `wd23-lad23-ctyua23` is about the local authorities etc?

app/broadcast_areas/models.py
```diff
@@ -241,9 +241,14 @@ def __init__(self, row):
         self.name_singular = name_singular
         self.is_group = bool(is_group)
         self.items = BroadcastAreasRepository().get_all_areas_for_library(self.id) if self.id != "postcodes" else []
-        self.item_ids = []
-        for item in self.items:
-            self.item_ids.append(item[0])
+        self.item_ids = (
```
gds-gharper: I'm a little concerned that anything which loads the broadcast_areas model is going to incur the cost of loading all these things from the database.  Is it worthwhile profiling these or maybe even just lazy loading them when they're accessed?

app/main/forms.py
```diff
@@ -542,6 +543,7 @@ def __init__(
         library_ids=None,
         item,
         area_id_parser,
+        check_spelling=False,
```
gds-gharper: Do we really need an extra parameter just to tell it to add the string "please check spelling" to the error message?

Can't we just decide that based on the context of the page?

app/main/forms.py
```diff
@@ -2013,6 +2019,40 @@ def validate(self, extra_validators=None):
...
+    areas = GovukTextareaBulkField("", item="Local authority", area_id_parser=_parse_ids, check_spelling=True)
+
+    def validate(self, extra_validators=None):
+        valid = super().validate(extra_validators)
+
+        if not valid:
+            # Form-level error message based on field's error code, returned if validation files
+            error_messages = {
+                "missing_data": "Enter at least 1 local authority",
+                "duplicates": "All local authorities must be unique",
```
gds-gharper, line 2040: Does it matter if they put the same area in twice?  It'll only get included once anyway.

Perhaps warn them but not make it an errror if thats possible?

gds-gharper (changes requested):

app/broadcast_areas/models.py
gds-gharper: I'd be interested to see some profiling to see how long this data prefetching takes

app/broadcast_areas/models.py
leylayaltiligil: Its calling 3 separate `BroadcastAreasRepository` methods, but they are all similarly named - I will change this if you think they're _too_ similarly named. 
Or do you think we should just call one method once and then each time retrieve what's needed from those results for each of those attributes?

app/broadcast_areas/repo.py
leylayaltiligil: I've removed that comment as it wasn't relevant (leftover as I copied the query from elsewhere in the file and repurposed) and I've renamed the function to `get_all_area_names_and_ids_for_library` as it returns both names and IDs.

app/main/forms.py
leylayaltiligil: Okay I will get Mehmet's thoughts on this as errors were provided as part of the design that I've worked to, and this has been implemented as part of FWA form validation too. 
It makes sense - though unsure of where I'd render the warning so its still relevant for form input but also let the user still progress with the journey.

app/broadcast_areas/repo.py
gds-gharper: My query is about the 'lad23-' filter on the like, that seems very specific to local authorities and the function name is not.

app/broadcast_areas/models.py
gds-gharper: But both calls reference the same rows in the database, so you're fetching the same rows from the database multiple times, this is incredibly inefficient.  Why not fetch them all once and then pick out the data you need from each?
The previous code did exactly that, what you've done is change that so it makes separate calls to the database, in most cases the database calls are going to be the slowest parts of this.

app/broadcast_areas/repo.py
leylayaltiligil: I've renamed the function to reference the fact it is only used for Local Authorities, and I've adjusted the comments.

app/broadcast_areas/models.py
leylayaltiligil: I've now reverted the changes to how `item_ids` created.

app/main/forms.py
leylayaltiligil: I've spoken to Mehmet regarding this and he's said that providing the error helps users avoid issues where local authorities have been accidentally duplicated, e.g. the user is provided with a list of 20 local authorities, they copy and paste this list into the text area but change one local authority which ends up being duplicate, the count is still the same but one of local authorities is now wrong.

app/broadcast_areas/models.py
leylayaltiligil: Those attributes are now cached properties and after some profiling it looks like there are significantly fewer function calls and reduced time taken for relevant routes, when compared with main branch.

app/main/forms.py
leylayaltiligil: The form field doesn't have access to page context at present, this can easily be passed in, but this would surely be the same as passing in `check_spelling` parameter. Then we would need additional logic to determine based on page context whether or not to display the added string, is it not simpler to just check whether `check_spelling` is True or False?
Would it help if `check_spelling` were renamed to something like `display_check_spelling_message`?

gds-gharper: Couldn't the message just be the same, I'm not sure that one piece of extra text in that one message warrants adding an extra parameter to a bunch of functions

leylayaltiligil: I've removed the logic and the content, as per earlier discussion and following Designer's agreement.

spansh (dismissed):

gds-gharper (approved):

PR #390, opened 2026-05-12
EAS-2973: Update phone estimation content

leylayaltiligil:
This PR changes the displayed content regarding estimated mobile phone count and includes relevant adjustments to tests.

### DIFF
--- a/app/templates/components/alert-summary-list.html
+++ b/app/templates/components/alert-summary-list.html
@@ -94,10 +94,12 @@
         <p class="govuk-body">
             {% if count_of_phones == 0 %}
                 Unknown number of phones
-            {% elif count_of_phones == count_of_phones_likely %}
-                {{ count_of_phones|format_thousands }} phones estimated
+            {% elif count_of_phones_likely < 900000 %}
+                Less than 1 million phones estimated
+            {% elif 900000 <= count_of_phones_likely < 1000000 %}
+                Just under 1 million phones estimated
             {% else %}
-                {{ count_of_phones|format_thousands }} to {{ count_of_phones_likely|format_thousands }} phones
+                More than 1 million phones estimated
             {% endif %}
         </p>
     {% endset %}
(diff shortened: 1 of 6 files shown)


app/templates/components/alert-summary-list.html
```diff
@@ -94,10 +94,12 @@ <h2 id="enlarged-map-dialog-message" class="govuk-visually-hidden screenreader-c
         <p class="govuk-body">
             {% if count_of_phones == 0 %}
                 Unknown number of phones
-            {% elif count_of_phones == count_of_phones_likely %}
-                {{ count_of_phones|format_thousands }} phones estimated
+            {% elif count_of_phones_likely < 900000 %}
+                Less than 1 million phones estimated
+            {% elif 900000 <= count_of_phones_likely < 1000000 %}
```
gds-gharper: Rather than have these magic numbers here, would it be better to put them into config/constants somewhere (and potentially have a function to handle them), something like
```
            {% alert_requires_senior_approval %}
            
                More than 1 million phones estimated
            {% elif alert_close_to_senior_approval %}
                Close to 1 million phones estimated
            {% else %}
                Less than 1 million phones estimated
            {% endif %}
```

app/templates/views/broadcast/macros/area-map.html
```diff
@@ -45,10 +45,12 @@
...
-        {{ count_of_phones|format_thousands }} phones estimated
+      {% elif count_of_phones_likely < 900000 %}
+        Less than 1 million phones estimated
+      {% elif 900000 <= count_of_phones_likely < 1000000 %}
+        Just under 1 million phones estimated
       {% else %}
-        {{ count_of_phones|format_thousands }} to {{ count_of_phones_likely|format_thousands }} phones
+        More than 1 million phones estimated
       {% endif %}
     </p>
```
gds-gharper, line 48: Would it make sense to move this block (which is repeated below) into a component?

gds-gharper (changes requested):

gds-gharper (dismissed):
I still think the magic numbers could be moved out of the templates (they're business logic) but at least they're now only in one place.

leylayaltiligil:
> I still think the magic numbers could be moved out of the templates (they're business logic) but at least they're now only in one place.

Ah yep, I will move the numbers

gds-gharper (approved):

PR #394, opened 2026-05-21
EAS-2955: create outbound email functionality

Dave-Johnson-GDS:


app/main/forms.py
```diff
@@ -2095,3 +2095,16 @@ class ChooseTemplateFieldsForm(StripWhitespaceForm):
         },
         validators=[DataRequired(message="Select which fields you'd like to use to populate template")],
     )
+
+
+class EmailSummaryForm(StripWhitespaceForm):
```
gds-gharper, line 2100: Instead of bundling all of these things which can be calculated in hidden fields, just send the alert GUID.  Then when you actually go to send the email calculate the fields.

In general if you can calculate something you shouldn't include it as a field.  This is even more important when it comes to security where you should never trust user input.  Just because something is a hidden field doesn't mean it can't be edited by a user.  The adage is "never trust user input", and you should always think "What would happen if someone edited this field to contain something nefarious".

In this case, if the user put in an invalid alert GUID we can check whether the user has access to that alert and service and if not we can throw an error.

https://dev.to/_garybell/never-trust-user-input-4ff1

You could in theory put these fields into the stash rather than form fields, but I'm not a big fan of using the stash either, that would be safer from a security perspective, but you're still passing parameters in that you could easily calculate.

app/models/service.py
```diff
@@ -252,6 +252,10 @@ def able_to_accept_agreement(self):
             Organisation.TYPE_NHS_LOCAL,
         }
 
+    @property
+    def contact_emails(self):
+        return self._dict["email_addresses"]
```
gds-gharper: I wonder if we should use a more specific name for the attribute?  email_addresses is a pretty broad name, alert_notification_addresses perhaps?

app/notify_client/broadcast_message_api_client.py
```diff
@@ -121,5 +121,18 @@ def get_broadcast_provider_statuses(self, service_id, broadcast_message_id):
         """
         return self.get(f"/service/{service_id}/broadcast-message/{broadcast_message_id}/provider-statuses")
 
+    def send_alert_summary_email(self, service_id, broadcast_message_id, data):
```
gds-gharper: I think we want some documentation here, you're taking a very generic parameter name "data" and throwing it straight at the API.  There's no way for anyone reading this code to know what the data parameter should contain and we also don't do any sanity checking on the data ether.

In general I'd be tempted to break the actual data into actual keyword/named parameters and then combine them here.  That way the function is self documenting, you can also run sanity checks on the data if so desired.

gds-gharper (changes requested):

app/main/forms.py
Dave-Johnson-GDS: Thanks Gareth - EmailSummaryForm updated and all hidden fields removed.  Values are either retrieved from the broadcast_message object or passed directly to the page when rendering, instead of via the form.  The only thing being passed around now via the form is the user editable summary text.

app/models/service.py
Dave-Johnson-GDS: Makes sense - updated.

app/notify_client/broadcast_message_api_client.py
Dave-Johnson-GDS: Keyword/named parameters now passed through instead of the single data object.

gds-gharper (approved):

PR #397, opened 2026-05-26
Fix/Includes county and unitary authorities in local authority area lookup

leylayaltiligil:
This PR fixes an issue in which certain local authorities weren't being returned when name entered as part of a list. This is because the lookup list that the field validation checks against didn't include County and Unitary authorities, this was a mistake and they are now included.


jamesorlakin (dismissed):

jamesorlakin (dismissed):

gds-gharper (dismissed):

leylayaltiligil:
I've adjusted the existing test for this functionality to assert that both Local Authority District (`lad25-`) areas and County and Unitary authority (`ctyua25-`) areas can be added using the textarea, as has been requested.

gds-gharper (approved):

PR #413, opened 2026-06-23
EAS-3104 Display Local Authority > Electoral Ward hierarchy

jonathan-owens-gds:
This PR renders area labels with the format "Local Authority > Electoral Ward". The new CSS class is added to avoid the full width rendering when labels are too big to fit side by side, as shown in the screenshot below:

image


gds-gharper (changes requested):
Would it make sense to make the individual ward display a component, so we if we do want to change it in future we're only changing it in one place?

gds-gharper (approved):

PR #421, opened 2026-07-03
EAS-3124: Check folder access within template routes

leylayaltiligil:
This PR adds `checks_template_folder_permissions_if_exists` function and references in relevant views.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -34,11 +34,19 @@ form_objects = {
 }
 
 
-def checks_template_folder_permissions_if_exists(template_folder_id):
-    # Checks user has permissions for folder that template would be created in
-    if template_folder_id:
-        template_folder = current_service.get_template_folder(template_folder_id)
-        current_user.has_template_folder_permission_or_403(template_folder)
+def assert_template_folder_permissions(template_folder):
+    # Check the current user has permission for the given template folder.
+    # Returns True if no folder ID is provided (no permission check needed), OR
+    # the user has permission for the folder.
+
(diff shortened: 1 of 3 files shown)


app/models/user.py
```diff
@@ -266,6 +266,11 @@ def has_template_folder_permission(self, template_folder, service=None):
 
         return self.id in template_folder.get("users_with_permission", [])
 
+    def has_template_folder_permission_or_403(self, template_folder, service=None):
+        if not self.has_template_folder_permission(template_folder, service=service):
+            abort(403)
+        return True
+
```
jamesorlakin: I feel this breaks the separation a bit, `User` doesn't really represent anything web-based: the 403 logic could be a method elsewhere that then calls `has_template_folder_permission`

app/main/views/templates.py
```diff
@@ -34,13 +34,20 @@
 }
 
 
+def checks_template_folder_permissions_if_exists(template_folder_id):
```
jamesorlakin: This could be `assert_template_folder_permissions` and do the 403-ing instead of User

app/models/user.py
leylayaltiligil: I've removed it in favour of added [`assert_template_folder_permissions`](https://github.com/alphagov/emergency-alerts-admin/pull/421/changes#diff-ab6836e1de1ae35fdbf46e68b40e71ca658c734751950019b2c08aba9c4084a1R37#L37).

jamesorlakin (approved):

PR #427, opened 2026-07-13
EAS-3150: dss cand 031 update password reset api

Dave-Johnson-GDS:
Either pass the old password, or the password reset token through to the API when making an update_password API call.

Old password is passed to API when a user changes their password from the User Profile admin screen.

Token is passed to API when a user logs in from a password reset link.

Similar checks are then performed in the API to the admin application, to ensure update_password requests are valid.


gds-gharper (approved):

Dave-Johnson-GDS:
Note, functional tests have run successfully against my hosted dev env, with both API and ADMIN changes deployed.

